PR #1, opened 2021-03-14
add Rust programs

BurntSushi:
So I kind of ran away with this a bit. I wrote 5 different Rust programs. If you only want the "simple" and "optimized" versions, then that's OK, but I think the others are interesting:

* `simple` and `optimized` roughly mirror the Go programs by the same name.
* `optimized-customhashmap` is an approximate port of your optimized C program.
* `optimized-trie` was an experiment of mine to drop the use of a hash table and instead use a trie. Ultimately, this ended up being slower. Maybe twice as slow.
* `bonus` shows a Rust program that does full Unicode aware word segmentation. It's about as fast as the Rust and Go `simple` variants. It's a nice bonus challenge, because I wonder what this program would like ported to other languages. :-)

Here are some benchmarks I ran locally with `hyperfine`. Not that the timings in particular are too important, but it's a nice sanity check. If you get dramatically different results, then maybe it's worth looking more carefully.

```
$ hyperfine './rust/simple/target/release/countwords < kjvbible_x10.txt'
Benchmark #1: ./rust/simple/target/release/countwords < kjvbible_x10.txt
  Time (mean ± σ):      1.023 s ±  0.050 s    [User: 1.012 s, System: 0.010 s]
  Range (min … max):    0.949 s …  1.076 s    10 runs

$ hyperfine './rust/optimized/target/release/countwords < kjvbible_x10.txt'
Benchmark #1: ./rust/optimized/target/release/countwords < kjvbible_x10.txt
  Time (mean ± σ):     270.2 ms ±   6.9 ms    [User: 263.1 ms, System: 6.9 ms]
  Range (min … max):   261.5 ms … 282.0 ms    11 runs

$ hyperfine './rust/optimized-trie/target/release/countwords < kjvbible_x10.txt'
Benchmark #1: ./rust/optimized-trie/target/release/countwords < kjvbible_x10.txt
  Time (mean ± σ):     338.0 ms ±   5.8 ms    [User: 291.6 ms, System: 45.0 ms]
  Range (min … max):   324.2 ms … 343.6 ms    10 runs

$ hyperfine './rust/optimized-customhashmap/target/release/countwords < kjvbible_x10.txt'
Benchmark #1: ./rust/optimized-customhashmap/target/release/countwords < kjvbible_x10.txt
  Time (mean ± σ):     208.0 ms ±   7.3 ms    [User: 197.6 ms, System: 10.2 ms]
  Range (min … max):   186.4 ms … 216.1 ms    15 runs

$ hyperfine './rust/bonus/target/release/countwords < kjvbible_x10.txt'
Benchmark #1: ./rust/bonus/target/release/countwords < kjvbible_x10.txt
  Time (mean ± σ):      1.125 s ±  0.042 s    [User: 1.114 s, System: 0.010 s]
  Range (min … max):    1.072 s …  1.183 s    10 runs

$ hyperfine './simple-go < kjvbible_x10.txt'
Benchmark #1: ./simple-go < kjvbible_x10.txt
  Time (mean ± σ):     900.0 ms ±  35.4 ms    [User: 954.9 ms, System: 34.4 ms]
  Range (min … max):   841.7 ms … 948.3 ms    10 runs

$ hyperfine './optimized-go < kjvbible_x10.txt'
Benchmark #1: ./optimized-go < kjvbible_x10.txt
  Time (mean ± σ):     299.0 ms ±  25.9 ms    [User: 288.0 ms, System: 16.7 ms]
  Range (min … max):   265.1 ms … 361.3 ms    10 runs

$ hyperfine './simple-c < kjvbible_x10.txt'
Benchmark #1: ./simple-c < kjvbible_x10.txt
  Time (mean ± σ):     737.4 ms ±  30.9 ms    [User: 725.9 ms, System: 11.0 ms]
  Range (min … max):   690.1 ms … 779.8 ms    10 runs

$ hyperfine './optimized-c < kjvbible_x10.txt'
Benchmark #1: ./optimized-c < kjvbible_x10.txt
  Time (mean ± σ):     179.8 ms ±   4.0 ms    [User: 174.9 ms, System: 4.6 ms]
  Range (min … max):   166.5 ms … 185.0 ms    17 runs

$ hyperfine './optimized-c-llvm < kjvbible_x10.txt'
Benchmark #1: ./optimized-c-llvm < kjvbible_x10.txt
  Time (mean ± σ):     185.4 ms ±   7.4 ms    [User: 175.9 ms, System: 9.3 ms]
  Range (min … max):   167.4 ms … 191.8 ms    16 runs
```

### DIFF
--- a/rust/.gitignore
+++ b/rust/.gitignore
@@ -1,0 +1,2 @@
+/*/target
+/*/Cargo.lock
--- a/rust/bonus/Cargo.toml
+++ b/rust/bonus/Cargo.toml
@@ -1,0 +1,17 @@
+[package]
+name = "simple"
+version = "0.1.0"
+authors = ["Andrew Gallant <jamslam@gmail.com>"]
+edition = "2018"
+
+[[bin]]
+name = "countwords"
+path = "main.rs"
+
+[dependencies]
+anyhow = "1.0.38"
+bstr = "0.2.15"
+fxhash = "0.2.1"
+
+[profile.release]
+debug = true
--- a/rust/bonus/main.rs
+++ b/rust/bonus/main.rs
@@ -1,0 +1,73 @@
+// This is the Rust program I would write if someone asked for a *versatile*
+// program to find the frequency distribution of all the words, regardless of
+// case, in a stream. This handles any file that is at least conventionally
+// UTF-8 (so latin-1 is fine). It also uses the Unicode-aware word segmentation
+// algorithm and Unicode-aware case folding. This is despite NOT using Rust's
+// default Unicode string, since that requires the data to be valid UTF-8.
+//
+// It runs about as fast as the "simple" Rust and Go variants. We do two perf
+// tricks in this program: 1) we use bstr's `to_lowercase_into` to avoid
+// allocating a new string for every word, and 2) we look for an existing word
+// in our hashmap to avoid an allocation via the more idiomatic 'entry' API.
+// (These combined reflects a ~33% improvement in my ad hoc experiments.)
+// According to profiling, a bit over half the runtime is spent in word
+// segmenting. The rest seems somewhat evenly split between the actual
+// lowercasing and hashmap interactions.
+//
+// The gauntlet has been laid down. How hard is to port this program to other
+// languages? And when you do, what does its performance look like?
+
+use std::io::{self, Write};
+
+use bstr::{io::BufReadExt, BStr, BString, ByteSlice};
+use fxhash::{FxHashMap as HashMap};
+
+fn main() {
+    // Rust blocks the broken pipe signal by default, and instead returns it as
+    // an error from `write` if the consumer hangs up. So we look for it here
+    // and exit gracefully, as an end user would expect.
+    if let Err(err) = try_main() {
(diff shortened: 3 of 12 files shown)


benhoyt:
As mentioned in my email, thank you! This is great stuff.

pickfire:
@BurntSushi Would it be faster if we have another buffer to keep the hashmap keys for optimized? So instead of `HashMap<Vec<u8>, u64>`, we can have `HashMap<&[u8], u64>` and the `&[u8]` is pointing to the buffer in `Vec<u8>`, at least we don't get to have so many different buffers and we allocate lesser. Something like a bump allocator.

BurntSushi:
@pickfire Sounds like worth a try to me! I just didn't think of that. I think my _guess_ is that it wouldn't help. In the "optimized" Rust variant, we are already avoiding allocation in the common case. Your approach would save allocation in the case of adding a new word, but generally, it doesn't matter too much what you do in that case (in my experience writing these programs). But! I'm not certain. It's worth a shot IMO.

benhoyt:
Yeah, it's likely not to make much difference if it's only for the new-word case. But I'd accept a PR for `optimized.rs` if you want, @pickfire !

Issue #3, opened 2021-03-14
C# version


jftuga:
I have a C# version here: [freq.cs](https://github.com/jftuga/universe/blob/master/freq.cs)

With Visual Studio 2019, you can compile to `.exe` by running:

    C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\MSBuild\Current\Bin\Roslyn\csc.exe freq.cs

Note that this version [sorts the output](https://github.com/jftuga/universe/blob/master/freq.cs#L43) which you may not want in your benchmark version.  On line 45, you could change `tbl_sorted` to `tbl` to output unsorted results (after commenting out line 43).

**EDIT:** I just remembered this operates on a line-by-line basis and not word-by-word.

benhoyt:
Thanks for that. Yeah, we need to tokenize / parse the line into words. Also, it only needs to handle stdin and not files, so you could make that simplification. If you make those two changes, I'd link to it under Other Languages for the C# version.

I love how simple and clean the sorting is in C#.

Out of interest, how did you find out about this repository? From my comp.lang.forth posting?

jftuga:
I will work on this.

I found this because @BurntSushi cloned your repo - I guess to write the rust version.  I was looking at what else he has done, other than `ripgrep`.

benhoyt:
Nice, and no problems -- was just curious!

Thanks for the C# version. Ideally it'd run on Linux, so I can benchmark it against the other language versions on the same machine. I guess the best way to do that is Mono? https://www.mono-project.com/

jftuga:
I created a PR.  I would agree that Mono would be the best way to test.  I wrote this program using Windows 10 and Visual Studio 2019.  It's pretty simple, so I would assume Mono would not have any problems running it.

benhoyt:
Excellent, thanks. I'll close this issue in favour of your PR.

jbergens:
I have not checked the code yet but .NET on Linux is most often done by using .NET Core or the new .NET 5. Then you can both compile and run on Linux and the program will probably be faster.

benhoyt:
Yeah, I'm now running it with .NET 5 on Linux.

PR #5, opened 2021-03-14
created C# version

jftuga:
I have made the requested changes as discussed in [issue #3](https://github.com/benhoyt/countwords/issues/3).

### DIFF
--- a/simple.cs
+++ b/simple.cs
@@ -1,4 +1,5 @@
-﻿
+﻿// Original version by John Taylor
+
 using System;
 using System.IO;
 using System.Collections.Generic;
--- a/simple.cs
+++ b/simple.cs
@@ -1,0 +1,46 @@
+﻿
+// to compile:
+// csc.exe -o+ simple.cs
+
+using System;
+using System.IO;
+using System.Collections.Generic;
+using System.Collections.Concurrent;
+using System.Linq;
+
+namespace freq
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            Dictionary<string, int> tbl = new Dictionary<string, int>(StringComparer.Ordinal);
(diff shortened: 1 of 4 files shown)


simple.cs
```diff
@@ -0,0 +1,46 @@
...
+                words = line.Split((char[])null);
+                foreach (string w in words)
+                {
+                    if(w.Length == 0) {
+                        continue;
+                    }
+                    string lower = w.ToLower();
+                    tbl[lower] = (tbl.TryGetValue(lower, out value)) ? tbl[lower] = value + 1 : tbl[lower] = 1;
+                }
+            }
```
benhoyt: I'm not a C# programmer, but it seems to me there are a few things that aren't particularly simple (and hence likely not idiomatic) here. You can use `StringSplitOptions.RemoveEmptyEntries` on the `Split()` call to avoid the `w.Length` check. You can use `Dictionary.GetValueOrDefault` to simplify the inner loop. Also, it's a bit better more efficient (and no less simple) to lowercase the entire line at once, rather than per word.

What about the following (also tweaked a couple of the names):

```c#
        var counts = new Dictionary<string, int>();
        string line;
        while ((line = Console.ReadLine()) != null)
        {
            line = line.ToLower();
            var words = line.Split(' ', StringSplitOptions.RemoveEmptyEntries);
            foreach (string word in words)
            {
                counts[word] = counts.GetValueOrDefault(word, 0) + 1;
            }
        }
        var ordered = counts.OrderByDescending(pair => pair.Value);
        foreach (var entry in ordered)
        {
            Console.WriteLine("{0} {1}", entry.Key, entry.Value);
        }
```

jftuga: That looks better!

PR #7, opened 2021-03-15
Optimize simple.cs

Calytic:
The largest portion of the performance cost of this method is strictly using Console.ReadLine and Console.WriteLine (in particular Console.WriteLine 32187 times)

Instead, use a StringBuilder to combine the messages and call Console.Write once at the end.

In my testing, making the above change cut the average execution from 2.1 seconds to 0.5 seconds.

Even still, the largest portion of the performance cost is using Console. The word counting in isolation executes in 0.1 seconds on my machine.

Anyway, interesting project and keep up the good work.

### DIFF
--- a/simple.cs
+++ b/simple.cs
@@ -22,9 +22,11 @@ class Program
             }
         }
         var ordered = counts.OrderByDescending(pair => pair.Value);
+        StringBuilder stringBuilder = new StringBuilder();
         foreach (var entry in ordered)
         {
-            Console.WriteLine("{0} {1}", entry.Key, entry.Value);
+            stringBuilder.AppendLine($"{entry.Key} {entry.Value}");
         }
+        Console.Write(stringBuilder.ToString());
     }
 }


Calytic:
Btw, I want to add that other suggestions (from HN, not me) of..
1. Using a different collection (not Dictionary), 
2. Removing allocations in `foreach`, using `for` instead
3. Removing GetValueOrDefault and using a simple condition `if`

I tried all of those and they all had negligible impact in .NET 5.0. It's obvious to me that this experiment is optimizing for CPU cycles and not memory footprint. That might be an interesting addition, but it would probably be non-trivial compared to simply measuring execution time.

Calytic:
Further, I want to add that the immutable string implementation in C# was (and is?) highly contentious compared to other languages, and personally I would consider StringBuilder usage idiomatic in C# for this reason.

https://stackoverflow.com/questions/15894906/memory-allocation-in-string-concatenation-in-c-sharp

That said, I would understand if you wanted to split this up into the optimized version.

aloisdg:
Not optimize at all, but here is a Linq version for fun

```C#
using System;
using System.Collections.Generic;
using System.Linq;

class Program
{	
    static void Main(string[] args)
    {
		var pairs = ReadLines()
			.SelectMany(x => x)
			.GroupBy(x => x)
            		.Select(group => $"{group.Count()} {group.Key}")
			.OrderByDescending(x => x);

		Console.WriteLine(string.Join(Environment.NewLine, pairs));
    }
	
    static IEnumerable<string[]> ReadLines()
    {
        string line;
        while ((line = Console.ReadLine()) != null)
        {
            yield return line.ToLower().Split(' ');
        }
    }
}
```

[try it online](https://tio.run/##XVFdT8IwFH3efsWVmNgluj@A8AASooIaMfHB@FC262jStXjbAcuy3z5bxviwydJ7z/3YOaeJuUs0YdMURqgMFqWxmPfD8yweaykxsUIrE09RIYnkX8dMqN9@GCaSGwNvpDPieVgFIbhjLLcigY0WKcy5UMxYcqNf38ApM9G@pwqDYMMJ1lyQgQG8I0/dTjQscpUgXqAnMOeqZDsYDGHXwlPSxXp0wuDsHKdY5rt8x3Wv2sdOUKEsi2o45M9Y1r125SulSKPyAU2CKnU8u@VOXhCMnQVaYvxJwqIneBATP2knbKI2grTKUdn4Bbe@fttKity4J1WHF548TlSRI/GlxPvOleGF@NacTlPbA9IV@0dwuxISgTGPOus6it0WFkVwNQBVSHkyqLqwqhQoUyC0Ban98vhDz/QWiUXxYi2FZTdwE53@WB@01E1jVwg/WsPSPZ6//ecxl4c@FsaHfw)

simple.cs
```diff
@@ -22,9 +22,11 @@ static void Main(string[] args)
             }
         }
         var ordered = counts.OrderByDescending(pair => pair.Value);
+        StringBuilder stringBuilder = new StringBuilder();
         foreach (var entry in ordered)
         {
-            Console.WriteLine("{0} {1}", entry.Key, entry.Value);
+            stringBuilder.AppendLine($"{entry.Key} {entry.Value}");
```
calledude, line 28: Technically, behind the scenes, `entry.Value` will be boxed. Furthermore, this allocates an entirely new string where atleast 2/3rds of the result already is a string. Proper usage would be to do something like 
```cs
stringBuilder.Append(entry.Key).Append(" ").AppendLine(entry.Value)
```

akraines:
Use a case-insensitive string comparer for the dictionary so you can skip .ToLower() 
(I created a pull request for that #24 )

benhoyt:
Hmm, that's strange. I've tried pulling these changes (including @calledude's optimization), and I don't see any noticeable timing difference. I'm running it via Mono on Linux, so maybe that's why -- are you on Windows?

BTW, I needed a `using System.Text;` to make this compile.

benhoyt:
In addition, I'd rather than optimize the simple.cs version (though I'm happy to simplify and/or make more idiomatic if we can do that). For optimizations, I'd be happy to add an optimized version.

One other thing: I remember that console output can be super duper slow on Windows, so maybe that's why yours went so much slower with the console write inside the loop.

PR #8, opened 2021-03-15
Add table from your blog

kokizzu:
copy table from your blog


benhoyt:
Thanks. However, I'd prefer to link to the results in the article rather than copy them here, so they're only in one place. I'll add that link soon.

PR #10, opened 2021-03-15
Remove pointers

kokizzu:
no need to use pointer when not needed


benhoyt:
Thanks, but I actually discuss why I use `map[string]*int` in the article for the optimized Go version -- it is somewhat faster as you avoid a map write for words that already exist (which is the common case).

PR #12, opened 2021-03-15
Simple version Swift

damuellen:


mimi89999 (approved):

simple.swift
```diff
@@ -0,0 +1,21 @@
...
+  var counts = [String:Int]()
+
+  for word in words {
+    let canon = word.lowercased()
+    counts[canon, default: 0] += 1
+  }
+
+  let ordered = counts.sorted { $0.1 > $1.1 }
+
+  ordered.forEach { entry in
```
mimi89999, line 14: Why not `for entry in ordered {`?

simple.swift
```diff
@@ -0,0 +1,21 @@
+import Foundation
+
+let input = FileHandle.standardInput
+
+if let text = String(data: input.readDataToEndOfFile(), encoding: .utf8) {
```
vojtamolda: I think this line violates the constraint mentioned the _Problem Statements and Constrains_ section of the [blog post](https://benhoyt.com/writings/count-words/). Solutions aren't supposed to load the entire file into memory. Maximum allowed is a 64KB chunk.

simple.swift
damuellen: I like to use forEach when nothing is mutated in the loop.

simple.swift
damuellen: I just skimmed the blog post, changed the code accordingly.

damuellen:
Made a mistake.

mimi89999:
@damuellen You can squash your commits and reuse the same PR.

mimi89999:
I guess that you wanted to write:
```
var counts = [String:Int]()

while let line = readLine() {
  let words = line.split { $0.isWhitespace }

  for word in words {
    let canon = word.lowercased()
    counts[canon, default: 0] += 1
  }
}

let ordered = counts.sorted { $0.1 > $1.1 }

ordered.forEach { entry in
  print(entry.key, entry.value.description)
}
```

simple.swift
vojtamolda: 👍

PR #17, opened 2021-03-15
Improve code generating kjvbible_x10.txt

mimi89999:


benhoyt:
Fair enough, but I dislike programming in bash so tend to opt for KISS solutions, and push programming to Python etc. :-)

PR #19, opened 2021-03-15
simple perl version

rlonstein:


benhoyt:
Thanks for your contribution. I've decided to take https://github.com/benhoyt/countwords/pull/9 as he got in first! :-) And it's very similar code in any case.

PR #20, opened 2021-03-15
Added simple nim example code.

csterritt:


euantorano:
Alternative, un-benchmarked version:

```nim
from tables import newCountTable, inc, sort, pairs
from algorithm import SortOrder
from strutils import split, toLowerAscii

proc main() =
  var table = newCountTable[string]()

  for line in stdin.lines:
    for word in line.toLowerAscii().split(' '):
      if word.len() == 0: continue
      table.inc(word)

  table.sort(SortOrder.Descending)

  for k, v in table.pairs():
    echo k, ' ', v

when isMainModule:
  main()
```

csterritt:
@euantorano Thanks! That benchmarks at 0.56 seconds instead of the 0.73 of my version... I'll switch to that!

benhoyt:
Thanks folks. Wow, simple nim code is fast!

PR #22, opened 2021-03-15
Add ruby version

llimllib:
The simple version is by me, and the optimized version is a direct translation of the python.

It's important to know that `""` and `0` are truthy in ruby, so the `if !...` checks are correct in this version even though ruby returns `0` for an `rindex` where the last instance is the first character.

To profile, use `gem install profile`, then:

```
$ ruby -r profile optimized.rb < kjvbible.txt > /dev/null
  %   cumulative   self              self     total
 time   seconds   seconds    calls  ms/call  ms/call  name
 60.39    15.04     15.04   885507     0.02     0.02  nil#
 17.55    19.41      4.37       68    64.26   356.91  Array#each
  6.23    20.96      1.55   821133     0.00     0.00  Hash#[]
  6.21    22.51      1.55   821133     0.00     0.00  Hash#[]=
  5.81    23.96      1.45   821200     0.00     0.00  Integer#+
  0.85    24.17      0.21    32187     0.01     0.01  IO#puts
  0.84    24.38      0.21    32187     0.01     0.01  Kernel#puts
  0.72    24.56      0.18        1   179.53   455.24  Hash#each
  0.66    24.72      0.16       67     2.44     2.44  String#split
  0.24    24.78      0.06    32187     0.00     0.00  IO#write
  0.22    24.83      0.05    32187     0.00     0.00  Integer#to_s
  0.22    24.89      0.05    32187     0.00     0.00  Integer#-@
  0.01    24.89      0.00        1     3.47   458.71  Enumerable#sort_by
  0.01    24.90      0.00       67     0.05     0.05  String#downcase
  0.00    24.90      0.00       67     0.01     0.01  String#+
  0.00    24.90      0.00        1     0.64     0.64  TracePoint#enable
  0.00    24.90      0.00      134     0.00     0.00  String#[]
  0.00    24.90      0.00       68     0.01     0.01  IO#read
  0.00    24.90      0.00      135     0.00     0.00  BasicObject#!
  0.00    24.90      0.00       67     0.00     0.00  String#rindex
  0.00    24.90      0.00        1     0.02     0.02  Class#new
  0.00    24.90      0.00        2     0.00     0.00  IO#set_encoding
  0.00    24.90      0.00        1     0.00     0.00  TracePoint#disable
  0.00    24.90      0.00        1     0.00     0.00  Integer#*
  0.00    24.90      0.00        1     0.00     0.00  Monitor#mon_owned?
  0.00    24.90      0.00        1     0.00     0.00  Kernel#respond_to?
  0.00    24.90      0.00        1     0.00     0.00  Hash#initialize
  0.00    24.91      0.00        1     0.00 24906.31  #toplevel
```

The `#nil` call stands out to me, but to be honest I'm not sure exactly where that's coming from.

### DIFF
--- a/optimized.rb
+++ b/optimized.rb
@@ -1,0 +1,20 @@
+counts = Hash.new(0)
+bufsize = 64*1024
+remaining = ''
+while 1 do
+  chunk = STDIN.read(bufsize)
+  if !chunk
+    break
+  end
+  chunk = remaining + chunk
+  last_lf = chunk.rindex("\n")
+  if last_lf.nil?
+    remaining = ''
+  else
+    remaining = chunk[last_lf+1..]
+    chunk = chunk[..last_lf]
+  end
+  chunk.downcase.split.each{ |w| counts[w] += 1 }
+end
+
+counts.sort_by{|k,v| -v}.each{|k,v| puts "#{k} #{v}"}
--- a/optimized.rb
+++ b/optimized.rb
@@ -8,7 +8,7 @@ while 1 do
   end
   chunk = remaining + chunk
   last_lf = chunk.rindex("\n")
-  if last_lf.nil?
+  if !last_lf
     remaining = ''
   else
     remaining = chunk[last_lf+1..]
--- a/simple.rb
+++ b/simple.rb
@@ -1,4 +1,4 @@
 #!/usr/bin/env ruby
 counts = Hash.new(0)
-STDIN.readlines.map { |line| line.downcase.split.each{ |w| counts[w] += 1 }}
-counts.sort_by{|k,v| -v}.each{|k,v| puts "#{k} #{v}"}
+STDIN.readlines.map { |line| line.downcase.split.each { |w| counts[w] += 1 }}
+counts.sort_by { |k,v| -v }.each { |k,v| puts "#{k} #{v}" }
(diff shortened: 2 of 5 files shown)


nhh:
"" and 0 are truthy 👍

simple.rb
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env ruby
+counts = Hash.new(0)
```
nhh, line 2: This is good stuff! Using `Hash.new(0)` would remove a null check in my version! 👍

llimllib:
@nhh I meant truthy where I said falsy! sheesh, thanks for the catch

llimllib:
Also I stole `each_line` from your answer, the `map` was stupid when we don't need the result

benhoyt:
Thank you muchly!

optimized.rb
```diff
@@ -0,0 +1,20 @@
...
+  end
+  chunk = remaining + chunk
+  last_lf = chunk.rindex("\n")
+  if !last_lf
+    remaining = ''
+  else
+    remaining = chunk[last_lf+1..]
+    chunk = chunk[..last_lf]
+  end
+  chunk.downcase.split.each{ |w| counts[w] += 1 }
```
nhh, line 17: I struggle a little about this block invocation with `#each`. It should be more optimized without a block being instantiated but I dont know if this is possible.

PR #24, opened 2021-03-15
Update simple.cs

akraines:
Use StringComparer.OrdinalIgnoreCase to avoid needing to call ToLower on every line

### DIFF
--- a/simple.cs
+++ b/simple.cs
@@ -10,11 +10,10 @@ class Program
 {
     static void Main(string[] args)
     {
-        var counts = new Dictionary<string, int>();
+        var counts = new Dictionary<string, int>(StringComparer.OrdinalIgnoreCase);
         string line;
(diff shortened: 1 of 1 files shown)


euantorano:
Note: I tried this change and while it works, it does mean that the output is different to all other implementations - this causes `test.sh` to fail.

akraines:
you can call to lower on output - it is still doing less work

akraines:
Did it improve the bentchmark?

benhoyt:
Yeah, I'd need to see a benchmark (with the output lowercased so the test succeeds). However, I'd like to keep the simple version as simple as possible, so if you do anything further here, please submit a new PR that adds an `optimized.cs` version. That said, for an optimized version like the Go version you'd probably need to do a lot more, like reading in chunks, custom tokenization/lowercase, etc.

PR #26, opened 2021-03-15
Add Julia simple version

alemelis:
here's a benchmark not including Julia startup overhead

```
julia> @benchmark wc(readlines("kjvbible_x10.txt")) |> sortdict
BenchmarkTools.Trial: 
  memory estimate:  1.57 GiB
  allocs estimate:  29217782
  --------------
  minimum time:     3.387 s (6.13% GC)
  median time:      3.419 s (7.87% GC)
  mean time:        3.419 s (7.87% GC)
  maximum time:     3.451 s (9.57% GC)
  --------------
  samples:          2
  evals/sample:     1
```

discussion here
https://discourse.julialang.org/t/count-words-challenge/57221/5

### DIFF
--- a/benchmark.py
+++ b/benchmark.py
@@ -32,6 +32,7 @@ programs = [
     ('AWK', 'gawk -f simple.awk', 'mawk -f optimized.awk', 'optimized uses `mawk`'),
     ('Forth', '../gforth/gforth-fast simple.fs', '../gforth/gforth-fast optimized.fs', ''),
     ('Shell', 'bash simple.sh', 'bash optimized.sh', 'optimized does `LC_ALL=C sort -S 2G`'),
+    ('Julia', 'julia simple.jl', '', ''),
 ]
 
 times = []
--- a/benchmark.py
+++ b/benchmark.py
@@ -33,7 +33,7 @@ programs = [
     ('AWK', 'gawk -f simple.awk', 'mawk -f optimized.awk', 'optimized uses `mawk`'),
     ('Forth', '../gforth/gforth-fast simple.fs', '../gforth/gforth-fast optimized.fs', ''),
     ('Shell', 'bash simple.sh', 'bash optimized.sh', 'optimized does `LC_ALL=C sort -S 2G`'),
-    ('Julia', 'julia simple.jl', '', ''),
+    ('Julia', 'julia simple.jl', None, 'by Alessandro Melis'),
     ('Perl', 'perl simple.pl', None, 'by Charles Randall'),
     ('JavaScript', 'node ./simple', 'node ./optimized', 'by Dani Biro'),
     ('Nim', './simple-nim', None, 'by csterritt and euantorano'),
--- a/README.md
+++ b/README.md
@@ -17,5 +17,6 @@ Thanks to these contributors for additional language versions:
 * Ruby: [Bill Mill](https://github.com/llimllib), with input from [Niklas](https://github.com/nhh)
 * JavaScript: [Dani Biró](https://github.com/Daninet)
 * Nim: [csterritt](https://github.com/csterritt) and [euantorano](https://github.com/euantorano)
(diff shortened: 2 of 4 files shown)


benhoyt:
Thanks for this!

benchmark.py
```diff
@@ -33,9 +33,10 @@ def time_run(cmdline):
     ('AWK', 'gawk -f simple.awk', 'mawk -f optimized.awk', 'optimized uses `mawk`'),
     ('Forth', '../gforth/gforth-fast simple.fs', '../gforth/gforth-fast optimized.fs', ''),
     ('Shell', 'bash simple.sh', 'bash optimized.sh', 'optimized does `LC_ALL=C sort -S 2G`'),
+    ('Julia', 'julia simple.jl', None, 'by Alessandro Melis'),
```
stevengj, line 36: Note that this means that your timing includes startup time, which in the case of Julia includes JIT compilation.   Since your test problem is so small (my [optimized version](https://discourse.julialang.org/t/count-words-challenge/57221/16?u=stevengj) runs in 80ms on my machine), the compilation overhead is potentially significant.

(For long-running performance-critical problems, typically Julia is used by launching `julia` once and then running any loops inside Julia, so that one-time compilation overheads don't matter, as opposed to launching `julia` with a little script that runs in a fraction of a second many times.)

I would normally recommend doing the timing *inside* the language in question, and running the code at least twice (and [keeping only the smallest time](https://arxiv.org/abs/1608.04295)) to avoid timing one-time costs like JITs (not to mention other noise).

benhoyt:
Thanks @stevengj. Yeah, I'm aware, but I didn't see an easy way to do that in the current benchmark system. Additionally, the benchmarks for all the other "interpreted" languages like Python and Ruby include the startup time in the benchmark, so it seems fair. If you write a tool like this in Julia, presumably you'd incur the cost in real life too.

stevengj:
> benchmarks for all the other "interpreted" languages like Python and Ruby include the startup time in the benchmark,

Julia is not an interpreted language—effectively everything is compiled.  (It's just that the compiler executes at runtime.)

> If you write a tool like this in Julia, presumably you'd incur the cost in real life too.

In real life if you care about performance for a problem using Julia, it will be for a problem that takes more than a fraction of a second to run (or you are running it many times, in which case you would put the loop in Julia).

If you're running a big calculation for minutes or hours, startup time is irrelevant.  (Or if you're working interactively, you typically leave a session open in an environment like Jupyter for minutes or hours.)

benhoyt:
Fair enough. I can't think of a fair way to subtract startup overhead from Julia without doing it for all the compiled-on-the-fly languages as well, so it gets messy fast.

stevengj:
Every language you are benchmarking provides timer facilities.  Why not have them print out the elapsed time as the first line of their output?

stevengj:
Either that or make the problem bigger — e.g. have every program process the file 100 times.

BurntSushi:
A simpler solution would be to just remove Julia from the benchmark. Its startup time makes it inappropriate for programs like this. Making the benchmark take 100 times as long to run just to accommodate Julia seems kind of ridiculous. And re-tooling measurement to be within each program is tedious.

benhoyt:
I agree with @BurntSushi - I'd rather not make the benchmark take 10x or 100x as long, and I don't have time to re-tool all the benchmarks to be within the program. Also, startup time does seem relevant here. For now, I've removed the Julia benchmark. Happy to re-add it later if we can find a simple workaround or way to speed up startup time.

stevengj:
Yes, if you're interested in benchmarking startup time, then any language with a JIT compiler (Julia, PyPy, Lua, …) is going to be at a big disadvantage for running a script that takes a fraction of a second and it's better not to include it.

(Personally I think benchmarking is more practically interesting when it's applicable to long-running tasks, but you have to choose what you want to measure.)

PR #27, opened 2021-03-15
Optimized C# version

euantorano:
This is a lightly optimized C# version, which makes a few changes:

- Use a `StreamReader` to read from stdin, with an input buffer size of 64K
- Use a `StreamWriter` to write results to stdout, which buffers output
- Rather than splitting the line into an array, use `Span<char>` instead to slice it up
- Build the project as a ready to run executable, which should save on startup time

This assumes the project is being built on and for Linux x64, and assumes a .net core 5 SDK is installed.

Results from testing locally:

| Language| Simple | Optimized
| -------- | ------ | ---------
| C# | 17.18 | 3.63


erikbozic:
Hi,
I was just looking into this and I believe the difference is mostly due to running netcore5.0 vs mono?
If I copy the simple.cs code into Program.cs the result is almost the same. Can you verify?

In that case maybe we should have a different benchmark for each runtime e.g.:
 - csharp-dotnet
 - csharp-mono
 
I could provide a Dockerfile for each so there's no need to install anything on the host machine.

euantorano:
> Hi,
> I was just looking into this and I believe the difference is mostly due to running netcore5.0 vs mono?
> If I copy the simple.cs code into Program.cs the result is almost the same. Can you verify?

Yes, that sounds correct. .NET 5 has some big performance improvements (as did .net core 3.1 before it).

> In that case maybe we should have a different benchmark for each runtime e.g.:
> 
> * csharp-dotnet
> * csharp-mono
> 
> I could provide a Dockerfile for each so there's no need to install anything on the host machine.

Makes sense to me.

akraines:
Why not use a OrdinalCaseInsensitive comparer (since you are anyway using a custom comparer) and only convert to lower when displaying the output?

benhoyt:
Unfortunately I can't seem to compile this (you're using a newer C# compiler? note that I'm running Mono on Linux). I've saved your source to `optimized.cs` and I get:

```
$ csc -optimize -out:optimized-cs optimized.cs
Microsoft (R) Visual C# Compiler version 3.6.0-4.20224.5 (ec77c100)
Copyright (C) Microsoft Corporation. All rights reserved.

optimized.cs(13,72): error CS8124: Tuple must contain at least two elements.
optimized.cs(13,73): error CS1526: A new expression requires (), [], or {} after type
optimized.cs(16,89): error CS1026: ) expected
optimized.cs(16,89): error CS1526: A new expression requires (), [], or {} after type
optimized.cs(16,89): error CS1002: ; expected
optimized.cs(16,89): error CS1513: } expected
optimized.cs(16,100): error CS1002: ; expected
optimized.cs(16,100): error CS1513: } expected
optimized.cs(50,80): error CS1526: A new expression requires (), [], or {} after type
```

benhoyt:
Closing this PR in favor of https://github.com/benhoyt/countwords/pull/32, which does seem to speed things up.

euantorano:
> Unfortunately I can't seem to compile this (you're using a newer C# compiler? note that I'm running Mono on Linux). I've saved your source to `optimized.cs` and I get:
> 
> ```
> $ csc -optimize -out:optimized-cs optimized.cs
> Microsoft (R) Visual C# Compiler version 3.6.0-4.20224.5 (ec77c100)
> Copyright (C) Microsoft Corporation. All rights reserved.
> 
> optimized.cs(13,72): error CS8124: Tuple must contain at least two elements.
> optimized.cs(13,73): error CS1526: A new expression requires (), [], or {} after type
> optimized.cs(16,89): error CS1026: ) expected
> optimized.cs(16,89): error CS1526: A new expression requires (), [], or {} after type
> optimized.cs(16,89): error CS1002: ; expected
> optimized.cs(16,89): error CS1513: } expected
> optimized.cs(16,100): error CS1002: ; expected
> optimized.cs(16,100): error CS1513: } expected
> optimized.cs(50,80): error CS1526: A new expression requires (), [], or {} after type
> ```

Yes, this is using .net 5 specific functionality and syntax, so you'd need to install the actual net stack rather than mono (.net is also available on linux now).

PR #29, opened 2021-03-15
Fixed printing of non null terminated char array

ansingh:
The ordered[i].word in c optimized version isn't null terminated, printing it with %s causes ub.


benhoyt:
Good spotting, thank you!

PR #30, opened 2021-03-15
Optimize the C++ version somewhat

jpakkane:
Here is an optimized version of the C++ code. The main reason it was slow was the same as for C: the standard library functions for parsing data are fairly slow.

I also updated the code to optionally use Abseil's fast hash map, as the one in the standard library is unfortunately known to be slow. Abseil is a bit of a pain to use as a dependency so it can't be used in the test directly as it just compiles things with a single `g++` invocation.

The timings for optimized builds on my machine are as follows:

- C: 0.15s
- C++: 0.37s
- C++ with Abseil: 0.29s
- C++ with Abseil and LTO: 0.25s
- C++ with Abseil and Clang: 0.24s

 I did not try Clang + LTO because the self-built Clang trunk I had was wonky for some reason.

### DIFF
--- a/optimized.cpp
+++ b/optimized.cpp
@@ -1,26 +1,71 @@
-// NOTE: this isn't really optimized C++ (as C++ is not my forte).
-// See the C version for more optimization.
-
 #include <algorithm>
 #include <cctype>
-#include <iostream>
 #include <string>
 #include <unordered_map>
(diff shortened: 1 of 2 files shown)


benhoyt:
Nice, thanks for this! I'll take a look and merge (probably later today).

PR #31, opened 2021-03-15
Add Lua version

ghost:
Lua is a versatile language. It's also very fast for a dynamic language. It's a simple program and can be improved, because right now I'm coping the table to another and allocate memory as well 2 times. 
`grep`        |   0.04 |      0.04 | `grep` baseline; optimized sets `LC_ALL=C`
`wc -w`       |   0.25 |      0.25 | `wc` baseline; optimized sets `LC_ALL=C`
C             |   0.75 |      0.16 | 
Rust A        |   1.08 |      0.27 | by Andrew Gallant
Lua           |   1.21 |           | 
Rust B        |   1.23 |      0.21 | also by Andrew: bonus and custom hash
C++           |   1.44 |      0.79 | "optimized" isn't really optimized
Perl          |   1.55 |           | by Charles Randall
Python        |   1.65 |      0.98 | 
Go            |   1.76 |      0.31 | 
AWK           |   2.00 |       | optimized uses `mawk`
Shell         |   2.16 |      1.70 | optimized does `LC_ALL=C sort -S 2G`

These are the results from my benchmark.


benhoyt:
I like this a lot, thanks! However, this looks more like the optimized version, as it reads in large chunks and then shuffles down the "rest". Could you make a `simple.lua` that's as simple as possible, and then move this version to `optimized.lua`?

Also, you probably want to lowercase the whole buffer, rather than word at a time.

Could you also please update test.sh/benchmark.py, and add credits to the README (you may need to merge the latest master).

catwell:
Here is a simple Lua version with basically the same performance as Python:

```lua
local counts = {}
local function insert(s) counts[s] = (counts[s] or 0) + 1 end
string.gsub(io.read("a"):lower(), "([^%s]+)", insert)

local t = {}
for k in pairs(counts) do table.insert(t, k) end
table.sort(t, function(a, b) return counts[a] > counts[b] end)

for _, v in ipairs(t) do print(v .. ' ' .. counts[v]) end
```

ghost:
I added simple and changed the optimized version. Thanks to catwell for the simple version. Added credits as well.

benhoyt:
I like that simple version, @catwell -- that looks great. However, there's something wrong with this PR, as it has 84 commits attached and a whole bunch of other stuff. Can you please rebase off the latest master and squash into a single commit?

PR #33, opened 2021-03-15
Add Kotlin simple tokenizer version and slightly optimised version

romanthekat:
Naive approach somewhat slower than using `StringTokenizer`, and closer to other languages stdlib simple solutions.
Bonus version uses regex :), but noticeably increases performance .

### DIFF
--- a/benchmark.py
+++ b/benchmark.py
@@ -47,9 +47,10 @@ programs = [
 times = []
 for program in programs:
     lang, simple, optimized, _ = program
-    print('Timing', lang)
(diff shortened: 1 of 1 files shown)


romanthekat:
possibly worth removing Gradle and rely purely on Kotlin compiler (though for Rust we used cargo)

benhoyt:
Thanks for this! Yeah, would be nice to remove the extra stuff and merge with latest master (to avoid conflicts).

PR #34, opened 2021-03-15
Add simple Nim implementation

indiscipline:
A simplest implementation with minimal allocations.
Some may find my usage of parens excessive for Nim, this is a matter of preference.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,5 +17,6 @@ Thanks to these contributors for additional language versions:
 * Ruby: [Bill Mill](https://github.com/llimllib), with input from [Niklas](https://github.com/nhh)
 * JavaScript: [Dani Biró](https://github.com/Daninet)
 * Nim: [csterritt](https://github.com/csterritt) and [euantorano](https://github.com/euantorano)
+* Julia: [Alessandro Melis](https://github.com/alemelis)
 
 See other versions [on Rosetta Code](https://rosettacode.org/wiki/Word_frequency).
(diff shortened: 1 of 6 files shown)


indiscipline:
Wow, there's already a Nim version. I'm so sorry, I needed to refresh the page before opening this PR.
However, my version is simpler while being effectively the same (although, I'm not sure about the speed effect of ordering of splitting and lowercasing).

benhoyt:
Yeah, we just added a nim version. Yours does look simpler, which is what we're aiming for in `simple.nim`. Feel free to update this PR off the latest master to change to your simplified version.

ZoomRmc:
Updated the PR

indiscipline:
Updated the PR.

euantorano:
Note:

The current Nim version has some changes in comparison to this version:

- Using a `main()` proc - the Nim compiler can sometimes optimise better when code is within a procedure than it can when in global scope
- Explicit imports rather than implicit - this is purely a matter of preference
- Explicit sorting when ordering - again, a matter of preference

Otherwise, the code is pretty similar which is nice to see - the Nim standard library is pretty complete and offers simple and obvious ways to accomplish tasks like this :)

It also looks like there are some unrelated changes in this PR with some Julia code too.

indiscipline:
Yes, using the `main()` proc is probably the correct decision, even if it may not bring any benefits in this case. Explicit imports are also have become the recommended way since not so long ago, rightly so. Unrelated commits are the result of syncing the branch, I probably should have done something else.

I'm absolutely positive there weren't any open or closed PRs with Nim versions nor `.nim` files in master branch at the time of me making the fork, don't understand how I could have missed it. Feel free to close this PR.

benhoyt:
Yeah, I think I'll close this PR in favor of the other one. If you or someone else wants to do another PR with an `optimized.nim` at a later date, I'll happily consider that. Thanks for your contribution!

PR #35, opened 2021-03-15
Seperate mono and netcore runtimes for csharp

erikbozic:
Hi,
I noticed there was only the mono benchmark present but no .NET Core version.  

C# code can be compiled and then ran on different runtimes (Mono, .NET Core, .NET Framework, etc..) with varying characteristics.  

Mono is the open source implementation, started when there was no cross-plat support from Microsoft.
However now also .NET Core exists, which is the Microsoft provided open source and cross-platform runtime. 

This is why I created a csharp folder in the root of the repo (same as rust), and have a different set of benchmarks for mono and .net core.  
Currently I only copied the simple implementation from mono to .net core and gave links to where to install each of the runtimes.

Here's the results for what I can easily run on my machine:
Nim           |   0.70 |           | by csterritt and euantorano
Go            |   1.06 |      0.38 | 
C#-dotnet     |   1.37 |           | 
Python        |   1.86 |      1.16 | 
JavaScript    |   2.21 |      1.30 | by Dani Biro
C#-mono       |   2.85 |           | original by John Taylor

I didn't make any effort to optimize anything in this PR, just to set it up so there can be multiple runtimes for .NET.  

I also tried a dockerized version of the benchmarks, but got 2x slower performance. I'm guessing it's something to do with the redirection of the input file, but I'm not sure.


erikbozic:
Similar was done in https://github.com/benhoyt/countwords/pull/32, closing.

PR #36, opened 2021-03-15
Very simple D implementation

dejlek:
All credits go to Herringway@IRC who actually did the implementation.


benhoyt:
Thank you! Are there any differences between optimized.d and simple.d? If not, let's just add only the simple.d version. Can you also please add to test.sh and update README.md with credits? Thanks!

PR #39, opened 2021-03-15
Simple PHP solution

MaxSem:


benhoyt:
Thanks! I had to tweak it a little to fix it counting empty words (the test didn't pass for me). I'll update the benchmarks a bit later.

PR #40, opened 2021-03-15
Update of the C++ optimized version in Modern C++: 

adevress:
Update of the C++ optimized version in Modern C++:

- Represent more a Modern C++ usage
- Use string_view and streams
- Avoid pointers arithmetic and C-style (fread/fwrite)
-  < 0.36s on laptop.

### DIFF
--- a/optimized.cpp
+++ b/optimized.cpp
@@ -1,79 +1,70 @@
 #include <algorithm>
 #include <cctype>
+#include <iostream>
 #include <string>
+#include <string_view>
 #include <unordered_map>
 #include <vector>
-#include <iostream>
-#include <cassert>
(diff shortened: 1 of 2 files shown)


benhoyt:
Thanks, this looks like a nice simplification / improvement.

ansingh:
Unfortunately, the behavior of cin.readsome is implementation specific. On OSX(Apple clang version 12.0.0) for instance this gives out empty output.

You can switch to cin.read(..), and then use cin.gcount() for the number of characters read. However, the majority of time is still spent on input on my machine (>2 sec). Switching to fread for the input speeds things up considerably (~400ms).

PR #41, opened 2021-03-16
Update simple.cpp

NexWeb:
The whole vector block is unnecessary.

### DIFF
--- a/simple.cpp
+++ b/simple.cpp
@@ -3,7 +3,7 @@
 #include <iostream>
 #include <string>
 #include <unordered_map>
-#include <vector>
+
 
 int main() {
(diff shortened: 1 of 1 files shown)


benhoyt:
No -- the problem statement says we have to sort the output most frequent first, and your change also removes the sorting. If you run `./test.sh` with your change, you'll see it fail.

NexWeb:
The output is exactly the same.  
count_words

benhoyt:
I think that is only by chance on the small test input. Try it again using `kjvbible_x10.txt` (see test.sh). Note that the container is `unordered_map`, so iteration order is undefined.

PR #42, opened 2021-03-16
better C++

ncm:
This version seems to be rather a lot faster.

### DIFF
--- a/optimized.cpp
+++ b/optimized.cpp
@@ -1,34 +1,35 @@
-// NOTE: this isn't really optimized C++ (as C++ is not my forte).
-// See the C version for more optimization.
-
 #include <algorithm>
 #include <cctype>
 #include <iostream>
+#include <iterator>
 #include <string>
 #include <unordered_map>
 #include <vector>
 
 int main() {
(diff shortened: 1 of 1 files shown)


benhoyt:
Thanks. However, note that `optimized.cpp` has already been updated a couple of times. Is this version faster than what's there now? (I'd be a bit surprised.)

ncm:
Eh, looks about the same.

But no need for std::ios::sync_with_stdio().

ncm:
BTW, unnecessarily-postfix op++ is non-idiomatic C++.

ncm:
When I cloned a few minutes ago, I got one version, but when I look on github.com/benhoyt, I see a different version. Maybe you had better measure on your hardware and see.

benhoyt:
Just checked, and this version takes ~1.6 seconds, whereas the existing master version takes ~0.46 seconds, so what we have now is a clear win. I did update the increment from `C++` to `++C`, though. :-) https://github.com/benhoyt/countwords/commit/61892e190608a2bae4f3c525417a45b8595c2af0

ncm:
I have no clue how you got it to run so slow.

PR #44, opened 2021-03-16
Add non-locale case conversion and minor improvement for C#

osman-turan:
This PR is about some minor improvements for C#:

1. It uses non-locale case conversion by using `ToLowerInvariant` instead of `ToLower`.
2. It uses a pre-increment operator instead of an "increment-by-one" statement for more idiomatic C#
3. It removes the unused `simple.cs` file in the repository root


benhoyt:
Nice, thanks for this. I'll merge now and update the benchmarks a bit later.

PR #45, opened 2021-03-16
Add simple OCaml version

doesntgolf:
Hope I've set up the benchmark and test correctly.... I'll try to have a go at an optimized OCaml version when I have more time. Thanks for putting this together and sharing your post about it, Ben!

### DIFF
--- a/simple.ml
+++ b/simple.ml
@@ -1,0 +1,21 @@
+let () =
+  let tbl = Hashtbl.create 33_000 in
+  try
+    let rec loop () =
+      read_line ()
+      |> String.lowercase_ascii
+      |> String.split_on_char ' '
+      |> List.iter (fun word ->
+          if String.length word != 0 then
+            match (Hashtbl.find_opt tbl word) with
+            | Some curr -> Hashtbl.replace tbl word (curr+1)
+            | None      -> Hashtbl.add tbl word 1
+        );
+      loop ()
+    in
+    loop ()
+  with End_of_file ->
+    Hashtbl.to_seq tbl
+    |> List.of_seq
+    |> List.sort (fun (_, a) (_, b) -> b - a)
+    |> List.iter (fun (word, count) -> Printf.printf "%s %d\n" word count)
--- a/test.sh
+++ b/test.sh
@@ -157,3 +157,8 @@ git diff --exit-code output.txt
 echo PHP simple
 php simple.php <kjvbible_x10.txt | python3 normalize.py >output.txt
 git diff --exit-code output.txt
+
+echo OCaml simple
+ocamlopt.opt -O3 simple.ml -o simple-ml
+./simple-ml <kjvbible_x10.txt | python3 normalize.py >output.txt
+git diff --exit-code output.txt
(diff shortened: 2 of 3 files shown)


benhoyt:
Thank you! Wow, the OCaml compiler is fast! Unfortunately I get a stack overflow during execution when running against the test file. Any ideas?

```
$ ocaml simple.ml <kjvbible_x10.txt 
Stack overflow during evaluation (looping recursion?).
$ ./simple-ml <kjvbible_x10.txt 
Fatal error: exception Stack_overflow
```

doesntgolf:
Hmmm, I clearly should've tried it with the x10 file. Seems to be something with Hashtbl.replace, I think. I'll have to dig in some more tomorrow.

test.sh
```diff
@@ -157,3 +157,8 @@ git diff --exit-code output.txt
 echo PHP simple
 php simple.php <kjvbible_x10.txt | python3 normalize.py >output.txt
 git diff --exit-code output.txt
+
+echo OCaml simple
+ocamlopt simple.ml -o simple-ml
```
rbjorklin: You might want to use `ocamlopt.opt -O3` :)

simple.ml
```diff
@@ -0,0 +1,26 @@
+let tbl = Hashtbl.create 1000000 in
```
rbjorklin: Might be "cheating" but we know the table has 32187 entries so make it ~33k?

simple.ml
```diff
@@ -0,0 +1,26 @@
...
+
+let process line =
+  line
+  |> String.lowercase_ascii
+  |> String.split_on_char ' '
+  |> List.iter (fun word ->
+      if (String.length word != 0) then
```
mseri: You could also use `word != ""` here

simple.ml
```diff
@@ -0,0 +1,26 @@
...
+        match (Hashtbl.find_opt tbl word) with
+        | Some curr -> Hashtbl.replace tbl word (curr+1)
+        | None      -> Hashtbl.add tbl word 1
+    )
+in
+
+let rec loop () =
+  try
```
mseri: I think to avoid the stack overflow you should move the `try ... with` around the call on line 26 instead of nesting it inside the recursive function (alternatively you could also use an imperative while loop)

test.sh
doesntgolf: Good call, thanks!

simple.ml
doesntgolf: Might as well!

simple.ml
doesntgolf: I actually used `word != ""` at first, but it was letting an empty word of some sort slip through, and piping to `python3 normalized.py` (via test.sh) was throwing an exception because of it. I'm still not quite sure why it was slipping through.

simple.ml
doesntgolf: Nice, that was indeed the cause. Thanks @mseri!

simple.ml
mseri: Oh, that's interesting, I will look into it.

doesntgolf:
@benhoyt this seems to be handling the bigger file better now. Thanks @rbjorklin and @mseri!

benhoyt:
Thanks - merged! I'll update the results later today

PR #46, opened 2021-03-16
Add simple Kotlin version

romanthekat:
Simple Kotlin version, apart from https://github.com/benhoyt/countwords/pull/33, now uses one Kotlin file w/o Gradle for building/compiling.
Naive approach somewhat slower than using `StringTokenizer`, and tokenizer is closer to other languages stdlib simple solutions, so it was chosen.

p.s. thanks for a nice article ^__^


morisil:
Please take a look at my proposal #53 I believe it is using Kotlin in more idiomatic way.

romanthekat:
@morisil yeah, #53 looks really neat ^_^ 
Love streaming api. Even though decided not to use it here :)
More Kotlin-ish, less Java-ish

benhoyt:
Thanks for the contribution. Given what you guys have said, I think I will close this in favor of #53. Thanks anyway!

PR #48, opened 2021-03-16
Fix remaining chunk processing in optimized.py

### LINKED ISSUE #28: optimized.py is wrong if the last line doesn't end with a newline
This input `the fox\nfox` should print: `fox 2\nthe 1\n` yet it prints `the 1\nfox 1\n`.

Goes to show that chunks are tricky.

Voultapher:
Fixes https://github.com/benhoyt/countwords/issues/28


benhoyt:
Thanks!

PR #50, opened 2021-03-16
Add F# implementation

yuriyostapenko:
~The simple implementation is very slow, since it is using immutable Map to store counters, which is probably one of the slowest things you could possibly do. This is certainly not an attempt at "how fast can it go" :)~

EDIT: I reconsidered this, since using Map would not probably be considered a reasonable implementation for this task even in idiomatic F#.

Optimized version could, in theory, be identical or very close to C# in performance, but that is not what I tried to do here.


benhoyt:
Thanks again!

PR #51, opened 2021-03-16
Remove intermediate vars in simple Perl version

MasterDuke17:
Makes it a bit more similar to the other simple scripts. Also speeds it for me from ~1.4s to ~1.3s.


benhoyt:
Sure, I'll take it - thanks!

PR #53, opened 2021-03-16
even simpler Kotlin version

morisil:


morisil:
I believe it is more functional and idiomatic Kotlin version than #46

Nice project! Please also take a look at my project with similar intent:

https://github.com/xemantic/java-2-times-faster-than-c

simple.kt
```diff
@@ -0,0 +1,8 @@
+fun main() = generateSequence { readLine() }
+        .flatMap { it.toLowerCase().splitToSequence(' ') }
+        .filter { !it.isBlank() }
```
sfesenko: IMO, it's better to `.filter { it.isNotEmpty() }` or `.filterNot(String::isEmpty)`

Also you may redefine `readLine` as 
```kotlin
val readLine = System.`in`.reader().buffered(65535)::readLine
```
and get "optimized" version for free (~30% boost)

morisil: Regarding the first part, good point, changed to `it.isNotBlank()` slightly longer but cleaner, I believe `it.isNotEmpty()` wouldn't work in this particular case.

Regarding the second point. This was my initial version, but Kotlin != Java. The current code can be easily set up as a multiplatform Kotlin project and compiled also as JavaScript and native and benchmarked separately, as I am doing in this project:

https://github.com/xemantic/java-2-times-faster-than-c

If I was to provide highly performant Kotlin code, it would look differently and probably might be dependant on JVM. For the simple one I see no good reasons to.

sfesenko: > The current code can be easily set up as a multiplatform Kotlin project

Technically - yes, it's possible to compile current version with kotlin-native, but
 - it runs ~ 9 sec vs ~2 sec (jvm)
 - it produces different output

 So, it seems, you'll need different version for each platform in any case

benhoyt:
Thanks for the contribution! I like the code, very sweet and streamy. However, can we do it without Gradle and all the extra .bat files, etc? Using just `kotlinc` like https://kotlinlang.org/docs/command-line.html#create-and-run-an-application ?

Ideally we'd just have `/simple.kt` and update test.sh / benchmark.py.

romanthekat:
> Ideally we'd just have `/simple.kt` and update test.sh / benchmark.py.

@morisil feel free to copy `test.sh`/`benchmark.py` updates I did in #46 for building and running Kotlin file w/o Gradle.
Basically `kotlinc -include-runtime simple.kt -d kotlin_simple.jar`, and running with `kotlin ./kotlin_simple.jar`

simple.kt
morisil: > * it runs ~ 9 sec vs ~2 sec (jvm)

Hmm, for me it was something like 1.25s vs 1.36s, in favor of JVM. I was using 1.4.31 compiler, I wonder where does the difference come from?

> * it produces different output

This I haven't checked, in general all these scripts are producing different output because often there are several words with the same number of occurrences. There are no rules for specified for the order of them. This is the reason why all the results are being normalized before comparison.

morisil: I will try to compile with mimalloc and see if there is any improvement

morisil:
> Thanks for the contribution! I like the code, very sweet and streamy. However, can we do it without Gradle and all the extra .bat files, etc? Using just `kotlinc` like https://kotlinlang.org/docs/command-line.html#create-and-run-an-application ?
> 
> Ideally we'd just have `/simple.kt` and update test.sh / benchmark.py.

Yes, I will update the code. @EvilKhaosKat has great example how to do it and actually installation of the needed toolchain is relatively simple. I will also add tests of compiling it to native code - standard and with mimalloc.

simple.kt
morisil: > * it produces different output

Indeed, `readLine` in case of kotlin native seems to be dropping data :( I don't see any kotlin native alternative for standard input processing which is disappointing.

PR #55, opened 2021-03-16
Keep simple.js even simpler (and faster) by not using a Map

laubsauger:
It's more basic and ~15% faster


benhoyt:
I like it, thanks!

PR #56, opened 2021-03-16
Rust minor optimizations

wfxr:
cc @BurntSushi

I have made some small optimizations which slightly improve the performance on my machine:

Before:
```
`grep`        |   0.05 |      0.05 | `grep` baseline; optimized sets `LC_ALL=C`
`wc -w`       |   0.53 |      0.23 | `wc` baseline; optimized sets `LC_ALL=C`
Rust A        |   1.55 |      0.38 | by Andrew Gallant
Rust B        |   1.62 |      0.31 | also by Andrew: bonus and custom hash
```

After:
```
`grep`        |   0.05 |      0.05 | `grep` baseline; optimized sets `LC_ALL=C`
`wc -w`       |   0.54 |      0.23 | `wc` baseline; optimized sets `LC_ALL=C`
Rust A        |   1.52 |      0.38 | by Andrew Gallant
Rust B        |   1.61 |      0.29 | also by Andrew: bonus and custom hash
```

### DIFF
--- a/rust/optimized-customhashmap/main.rs
+++ b/rust/optimized-customhashmap/main.rs
@@ -49,7 +49,7 @@ fn try_main() -> Result<(), Box<dyn Error>> {
                 }
             } else {
                 // 0x20 (6th bit) is the only different bit between lowercase and uppercase
-                buf[i] |= 0x20;
+                buf[i] |= (buf[i].is_ascii_uppercase() as u8) << 5;
                 if start.is_none() {
                     start = Some(i);
                 }
--- a/rust/optimized-customhashmap/main.rs
+++ b/rust/optimized-customhashmap/main.rs
@@ -67,9 +67,9 @@ fn try_main() -> Result<(), Box<dyn Error>> {
     }
 
     let mut ordered = counts.into_counts();
-    ordered.sort_by(|&(_, cnt1), &(_, cnt2)| cnt1.cmp(&cnt2).reverse());
+    ordered.sort_unstable_by_key(|&(_, count)| count);
 
-    for (word, count) in ordered {
+    for (word, count) in ordered.into_iter().rev() {
         writeln!(io::stdout(), "{} {}", std::str::from_utf8(&word)?, count)?;
     }
     Ok(())
--- a/rust/optimized-customhashmap/main.rs
+++ b/rust/optimized-customhashmap/main.rs
@@ -48,9 +48,8 @@ fn try_main() -> Result<(), Box<dyn Error>> {
                     hash = FNV_OFFSET;
                 }
             } else {
-                if b'A' <= b && b <= b'Z' {
-                    buf[i] += b'a' - b'A';
-                }
+                // 0x20 (6th bit) is the only different bit between lowercase and uppercase
+                buf[i] |= 0x20;
                 if start.is_none() {
                     start = Some(i);
                 }
--- a/rust/optimized-trie/main.rs
+++ b/rust/optimized-trie/main.rs
@@ -64,11 +64,10 @@ fn try_main() -> Result<(), Box<dyn Error>> {
     }
 
     let mut ordered = counts.frequencies();
-    ordered.sort_by(|&(_, cnt1), &(_, cnt2)| cnt1.cmp(&cnt2).reverse());
+    ordered.sort_unstable_by_key(|&(_, count)| count);
 
-    let mut stdout = io::stdout();
-    for (word, count) in ordered {
-        writeln!(stdout, "{} {}", std::str::from_utf8(&word)?, count)?;
+    for (word, count) in ordered.into_iter().rev() {
+        writeln!(io::stdout(), "{} {}", std::str::from_utf8(&word)?, count)?;
     }
     Ok(())
 }
--- a/rust/simple/main.rs
+++ b/rust/simple/main.rs
@@ -44,7 +44,7 @@ fn try_main() -> Result<(), Box<dyn Error>> {
     }
 
     let mut ordered: Vec<_> = counts.into_iter().collect();
-    ordered.sort_unstable_by_key(|&(_, count)| count);
+    ordered.sort_by_key(|&(_, count)| count);
 
     for (word, count) in ordered.into_iter().rev() {
         writeln!(io::stdout(), "{} {}", word, count)?;
--- a/rust/simple/main.rs
+++ b/rust/simple/main.rs
@@ -32,10 +32,14 @@ fn main() {
 }
 
 fn try_main() -> Result<(), Box<dyn Error>> {
+    let stdin = io::stdin();
+    let stdin = stdin.lock();
     let mut counts: HashMap<String, u64> = HashMap::new();
-    for line in io::stdin().lock().lines() {
-        for word in line?.split_whitespace() {
-            *counts.entry(word.to_lowercase()).or_insert(0) += 1;
+    for result in stdin.lines() {
+        let line = result?;
+        for word in line.split_whitespace() {
+            let canon = word.to_lowercase();
+            *counts.entry(canon).or_insert(0) += 1;
         }
     }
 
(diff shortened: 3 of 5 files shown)


rust/simple/main.rs
```diff
@@ -32,20 +32,17 @@ fn main() {
...
+        for word in line?.split_whitespace() {
+            *counts.entry(word.to_lowercase()).or_insert(0) += 1;
         }
     }
 
-    let mut ordered: Vec<(String, u64)> = counts.into_iter().collect();
-    ordered.sort_by(|&(_, cnt1), &(_, cnt2)| cnt1.cmp(&cnt2).reverse());
-    for (word, count) in ordered {
+    let mut ordered: Vec<_> = counts.into_iter().collect();
+    ordered.sort_unstable_by_key(|&(_, count)| count);
```
BurntSushi: I kind of see `sort_untable_by_key` as an optimization, but it's debatable. I'll let @benhoyt decide.

As for sorting by key and printing in reverse, I do think that's a bit simpler and I don't think that's an optimization, so that seems fine to me I think.

rust/simple/main.rs
```diff
@@ -32,20 +32,17 @@ fn main() {
...
-    let stdin = stdin.lock();
     let mut counts: HashMap<String, u64> = HashMap::new();
-    for result in stdin.lines() {
-        let line = result?;
-        for word in line.split_whitespace() {
-            let canon = word.to_lowercase();
-            *counts.entry(canon).or_insert(0) += 1;
+    for line in io::stdin().lock().lines() {
+        for word in line?.split_whitespace() {
+            *counts.entry(word.to_lowercase()).or_insert(0) += 1;
```
BurntSushi: I am actually not a fan of `line?`. I purposely used `result` because I think that's clearer.

Squashing `stdin().lock()` down into fewer lines seems OK to me though.

rust/optimized-customhashmap/main.rs
```diff
@@ -48,9 +48,8 @@ fn try_main() -> Result<(), Box<dyn Error>> {
                     hash = FNV_OFFSET;
                 }
             } else {
-                if b'A' <= b && b <= b'Z' {
-                    buf[i] += b'a' - b'A';
-                }
+                // 0x20 (6th bit) is the only different bit between lowercase and uppercase
+                buf[i] |= 0x20;
```
BurntSushi: Thanks for trying this! Interestingly, I am actually able to observe a measurable improvement. Before:

```
$ hyperfine './rust/optimized-customhashmap/target/release/countwords < kjvbible_x10.txt'
Benchmark #1: ./rust/optimized-customhashmap/target/release/countwords < kjvbible_x10.txt
  Time (mean ± σ):     218.0 ms ±   9.9 ms    [User: 209.8 ms, System: 7.5 ms]
  Range (min … max):   210.1 ms … 244.0 ms    13 runs
```

After:

```
$ hyperfine './rust/optimized-customhashmap/target/release/countwords < kjvbible_x10.txt'
Benchmark #1: ./rust/optimized-customhashmap/target/release/countwords < kjvbible_x10.txt
  Time (mean ± σ):     201.1 ms ±   5.1 ms    [User: 193.6 ms, System: 7.3 ms]
  Range (min … max):   191.7 ms … 211.9 ms    15 runs
```

BurntSushi (changes requested):
Mostly LGTM. I requested a few changes.

rust/simple/main.rs
wfxr: I originally thought the reason why use stable sort here is the tests require this. But seems unstable sort can also pass the tests so I did this. If we should not use unstable sort here, should it be ok to replace `sort_untable_by_key` with `sort_by_key`? I think it's more clear and maybe a little more efficient than `sort_by`.

rust/simple/main.rs
wfxr: Make sense. Both styles are ok for me. I will revert this change.

rust/simple/main.rs
BurntSushi: `sort_by_key` is fine.

Keeping the unstable sort in the other optimized programs sounds good though. Assuming you can see a measurable difference.

rust/optimized-customhashmap/main.rs
wfxr: @BurntSushi  Also thank you for contributing so many great solutions with detailed descriptions. I learned a lot from it!

rust/simple/main.rs
wfxr: Thanks for your review. In the current small test set there is indeed no significant performance improvement using unstable sort. IMO if we don't really need stable sort, it certainly has no negative effects on performance, nor does it increase the complexity of the code. So let's see @benhoyt 's decision. If stable sort actually is required I would replace it with `sort_by_key` then.

rust/optimized-customhashmap/main.rs
benhoyt: Hmmm, I'm not sure about this, because it also modifies some ASCII 

* \[ to {
* \\ to \|
* \] to \}
* \^ to \~
* \_ to 0x7f

It *happens* to work for the test input, because the Bible has no `@` or `^` in it :-), but I think we should revert it so it has a chance of handling more general inputs (we're not stripping punctuation).

rust/simple/main.rs
benhoyt: I think it's okay with unstable, though you can argue it either way. 1) plain sort is shorter and since you don't care, simpler, or 2) it doesn't need to be stable, so in a sense unstable sort is a "simpler" fulfilling of the requirements. That said, I lean towards `sort_by_key` for simple.rs, so let's leave it at that.

rust/optimized-customhashmap/main.rs
BurntSushi: You might try this instead? Taken from std impl of `to_ascii_lowercase`:

```rust
buf[i] |= (buf[i].is_ascii_uppercase() as u8) << 5;
```

rust/simple/main.rs
wfxr: > That said, I lean towards sort_by_key for simple.rs, so let's leave it at that.

@benhoyt Updated in 3f81b8b. Also restore stable sort for the "bonus" solution.

BurntSushi: FWIW, I think an unstable sort in the bonus version is fair game. It isn't under any constraints really, and it has some minor optimizations in it already.

wfxr: > FWIW, I think an unstable sort in the bonus version is fair game. It isn't under any constraints really, and it has some minor optimizations in it already.

Make sense. Thanks for your quick reply. Updated in 7589c4a.

rust/optimized-customhashmap/main.rs
benhoyt: Let's just revert this now, or update to the `(buf[i].is_ascii_uppercase() as u8) << 5` method if that's faster.

wfxr: @benhoyt Updated to @BurntSushi suggestion: 
```rust
buf[i] |= (buf[i].is_ascii_uppercase() as u8) << 5;
```

benhoyt:
Thanks!

rust/optimized-trie/main.rs
```diff
@@ -64,11 +64,10 @@ fn try_main() -> Result<(), Box<dyn Error>> {
...
 
     let mut ordered = counts.frequencies();
-    ordered.sort_by(|&(_, cnt1), &(_, cnt2)| cnt1.cmp(&cnt2).reverse());
+    ordered.sort_unstable_by_key(|&(_, count)| count);
 
-    let mut stdout = io::stdout();
-    for (word, count) in ordered {
-        writeln!(stdout, "{} {}", std::str::from_utf8(&word)?, count)?;
+    for (word, count) in ordered.into_iter().rev() {
+        writeln!(io::stdout(), "{} {}", std::str::from_utf8(&word)?, count)?;
```
pickfire, line 70: This might make it slower. See https://github.com/benhoyt/countwords/pull/87 to reduce this regression by having a bufwriter.

PR #58, opened 2021-03-16
Improve optimized C# with buffered read

osman-turan:
This PR improves C# performance with buffered read and some other minor optimizations.

I ported mostly from optimized C. As a result, it looks very similar to C now 😄 

**Before**
| Language | Simple | Optimized |
| -------- | ------ | --------- |
| C#       | 1.44   | 1.14      |

**After**
| Language | Simple | Optimized |
| -------- | ------ | --------- |
| C#       | 1.44   | 0.83      |

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ For the latest results (run on my machine), see the article's [performance resul
 
 Thanks to these contributors for additional language versions:
 
-* C#: [John Taylor](https://github.com/jftuga) and [Yuriy Ostapenko](https://github.com/uncleyo)
+* C#: [John Taylor](https://github.com/jftuga), [Yuriy Ostapenko](https://github.com/uncleyo) and [Osman Turan](https://github.com/osman-turan)
 * C++ optimized version: [Jussi Pakkanen](https://github.com/jpakkane) and [Adev](https://github.com/adevress)
 * Crystal: [Andrea Manzini](https://github.com/ilmanzo)
 * D: [Ross Lonstein](https://github.com/rlonstein)
--- a/benchmark.py
+++ b/benchmark.py
@@ -29,7 +29,7 @@ programs = [
     ('C++', './simple-cpp', './optimized-cpp', 'optimized by Jussi Pakkanen'),
     ('Python', 'python3 simple.py', 'python3 optimized.py', ''),
     ('Ruby', 'ruby simple.rb', 'ruby optimized.rb', 'by Bill Mill'),
-    ('C#', './csharp/simple/bin/Release/net5.0/simple', './csharp/optimized/bin/Release/net5.0/optimized', 'by John Taylor and Yuriy Ostapenko'),
+    ('C#', './csharp/simple/bin/Release/net5.0/simple', './csharp/optimized/bin/Release/net5.0/optimized', 'by John Taylor, Yuriy Ostapenko, Osman Turan'),
     ('AWK', 'gawk -f simple.awk', 'mawk -f optimized.awk', 'optimized uses `mawk`'),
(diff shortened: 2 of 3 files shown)


benhoyt:
Thanks for the contribution! I'd be happy to merge this, given that it's significantly faster (though you're right, everything does end up looking a bit like C :-). However, I want to defer to the original authors/contributors as to whether we should for C#. Thoughts, @erikbozic, @uncleyo, @euantorano ?

euantorano:
Looks good to me, but I’d be interested to see how it performs against a higher level version using a tool like BenchmarkDotNet: https://github.com/dotnet/BenchmarkDotNet

yuriyostapenko:
**LGTM** (obviously now it needs rebasing).

@benhoyt, I think it meets the criteria for the code that you've set in the blog post.

Now, question is, what is this all supposed to eventually become? :)

Currently, we have simple and optimized. I assume, simple shows idiomatic code people would reasonably write in a non-critical path for a similar problem in that given language. Optimized, on the other hand, only has constraints about staying memory safe (for memory safe languages), single threaded and using standard library data structures. And IMO it's fine.

I would personally also like some middle ground mode where code looks like something optimized, but still idiomatic for the ecosystem. Where it's not the same "a bit like C", with only GC and compiler optimizations left to differentiate the implementations.
Example: Choosing to use `OrdinalIgnoreCase` key comparison is not exactly the same level of optimization, as doing custom stream tokenization and patching character bits. But I don't know how to clearly set the rules here...

---

Then there's question of environment setup, reproducibility of results and comparability. If we compare different stacks, it would be fair to have each of them set up using some sort of best practice native solution. For example, if I understand correctly, you installed dotnet SDK via `snap`, which is [known for notoriously terrible performance](https://snapcraft.io/blog/snap-speed-improvements-with-new-compression-algorithm). I do not have any dotnet-specific numbers to prove that snap performs worse in this particular scenario (I never use snap), just aware of its poor reputation in this regard.

erikbozic:
I think there's a big difference between the simple/naive implementation and the optimized C-like implementation. 
We kind of lose on a potentially interesting language-specific implementations. But it is faster, so purely performance-wise this one should be the optimized version.

I do have something in the works that's more idiomatic but also somewhat closes the gap between the two implementations (0.80 vs 0.76 on my machine), but it's not fully working yet. (https://github.com/erikbozic/countwords/commit/fe308d8b582c92f5f4a4a70c7bfa29e5a349c57a unfortunately, I don't have time to finish right now, but if someone wants to help feel free )

Maybe the question should be if the goal is to:
-  find out how much performance can a specific language squeeze out
-  have a showcase of different approaches for a language with varying performance characteristics

yuriyostapenko:
@erikbozic, I don't think this repo should be about the latter (showcase many different approaches) as that IMO would better be served by a dotnet-only repo with BenchmarkDotNet.

erikbozic:
Alright if that's the general consensus, then let's merge this PR :)

I'll still try to come up with a way to make it more c#-like, but not lose on perf. 

I agree that BenchmarkDotNet would be the proper way to benchmark dotnet implementations. But as you said, it may not be appropriate for this repo - we can have a branch in our forks maybe.

EDIT: I also tried running it on the native AOT .NET runtime (https://github.com/dotnet/runtimelab/tree/feature/NativeAOT) which shaves off another ~15%. I assume startup performance. But that involves installing some dependencies manually for now.

benhoyt:
Thanks for this. I decided to go ahead and merge! I'll update the benchmarks in the article shortly. For me it's about 10-15% faster.

PR #59, opened 2021-03-16
C# optimized: remove string split and lowercase

erikbozic:
Instead of using `String.Split()`, we manually loop and substring the correct part of the line.
Also we use `StringComparer.OrdinalIgnoreCase` as the Dictionary comparer, so we don't have to `.ToLower()` for each word.
We still call `ToLower()` at the end so the results are output correctly.

I also used the top-level-function syntax (wihout `class Program` and `public static void Main`), since it's a very small program. 

Optimized version is usually ~20% faster.

BEFORE:
Go            |   1.05 |      0.37 | 
C#            |   1.36 |      1.08 | by John Taylor and Yuriy Ostapenko
Python        |   1.83 |      1.13 | 
JavaScript    |   2.18 |      1.27 | by Dani Biro

AFTER:
Go            |   1.05 |      0.37 | 
C#            |   1.44 |      0.87 | by John Taylor and Yuriy Ostapenko
Python        |   1.99 |      1.19 | 
JavaScript    |   2.21 |      1.40 | by Dani Biro

### DIFF
--- a/csharp/optimized/Program.cs
+++ b/csharp/optimized/Program.cs
@@ -1,42 +1,42 @@
-// Original version by John Taylor
-
 using System;
 using System.IO;
 using System.Collections.Generic;
 using System.Linq;
+using System.Runtime.CompilerServices;
 
-class Program
+var wordCount = new Dictionary<string, StrongBox<int>>(StringComparer.OrdinalIgnoreCase);
+
+using var stream = Console.OpenStandardInput();
+using var streamReader = new StreamReader(stream, System.Text.Encoding.UTF8, false, 64 * 1024);
+
+string line;
+while ((line = streamReader.ReadLine()) != null)
 {
-    public sealed class Ref<T> {
-        public Ref(T initialValue) {
-            Value = initialValue;
-        }
-        public T Value { get; set; }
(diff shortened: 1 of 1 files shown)


yuriyostapenko:
Not sure if it will make a big difference, but reading ASCII and changing `ToLower()` to `ToLowerInvariant()` may save a few extra ticks here and there.

erikbozic:
> Not sure if it will make a big difference, but reading ASCII and changing `ToLower()` to `ToLowerInvariant()` may save a few extra ticks here and there.

Tried ASCII and `ToLowerInvariant()` with no observable difference. 
Decided to stick with UTF8 (maybe more correct, not sure).
And I used `ToLower()`, because of the usage of `StringComparer.OrdinalIgnoreCase` in the dictionary - so they both match.
`StringComparer.InvariantCultureIgnoreCase` tanks performance pretty bad for some reason.

But if you come to some other conslusions I'd be glad to change stuff. :)

benhoyt:
Thanks for the contribution. It is faster for me too, and looks like a reasonable change. I'll update the results in the article later today.

PR #63, opened 2021-03-16
go fast, take chances...

rlonstein:
squeeze more out at the expense of convenience/clarity

### DIFF
--- a/optimized.d
+++ b/optimized.d
@@ -1,5 +1,6 @@
 import std.stdio;
 import std.string;
+import ctype = core.stdc.ctype;
 import std.algorithm.iteration : each, splitter;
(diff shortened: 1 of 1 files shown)


benhoyt:
This doesn't seem to make a noticeable performance difference for me with the test input - what are the before/after timings you're seeing?

rlonstein:
It's small, tenths and hundredths, from 1.91s to 1.79s here. Though I'm running out of ideas/knowledge for optimizing D.

benhoyt:
Sorry, I re-run with the latest changes and it's not any different for me (on Ubuntu Linux 64-bit). I'm going to close seeing there won't be any difference in the benchmarks table. Thanks anyway!

PR #66, opened 2021-03-16
Add a Bash version

lollipopman:
Thanks for doing this Ben!


benhoyt:
I appreciate the contribution! However, I don't think I'll add this version to the benchmarks, because it takes over 2 minutes to complete (even the optimized version), so it doesn't really seem a useful language to be doing it in. I'll still merge the PR so the code is there, but without the benchmarks. Sorry about that!

lollipopman:
makes sense, thanks @benhoyt for looking at it!

PR #67, opened 2021-03-16
 Implement a simple Haskell solution

noahhaasis:


benhoyt:
Happy to include this if you add to the readme/test/benchmark (you can comment out the others temporarily if you don't want to run all the benchmarks locally to test). Does the Haskell code read the whole file at once, or line-by-line? The latter (or at least not reading the whole file at once) is required by the constraints.

benhoyt:
Just saw this interesting article "Beating C with 80 lines of Haskell: wc": https://chrispenner.ca/posts/wc

noahhaasis:
Nice blog post looks really interesting. Although I have to say there're a lot of really elegant implementations in low level languages here, so I might just use one of them if I'd care for the extra performance.

simple.hs
```diff
@@ -0,0 +1,15 @@
+module Main where
+
+import Data.Char (toLower)
+import Data.List (sortBy)
+import qualified Data.Map.Strict as M
+
+main :: IO ()
+main = (putStr . prettify . sortCount . wc) =<< getContents
+  where
+    sortCount = sortBy (\(_, a) (_, b) -> compare b a) . M.toList
```
maurges, line 10: Why not `sortOn (negate . snd)`?

simple.hs
```diff
@@ -0,0 +1,15 @@
...
+    prettify = unlines . fmap (\(w, c) -> w ++ " " ++ show c)
+
+wc :: String -> M.Map String Int
+wc = foldr insertWord M.empty . words
```
maurges, line 14: Why not `M.fromListWith (+) . flip zip [1,1..] . map toLower . words`? fromList is more idiomatic and should be faster than foldr

simple.hs
```diff
@@ -0,0 +1,15 @@
+module Main where
+
+import Data.Char (toLower)
```
maurges, line 3: Why not use `Data.Text`?

noahhaasis: The README said that only the stdlib is allowed. Data.Text isn't in base right? Otherwise I would have used it (or ByteString).

simple.hs
noahhaasis: Because that's the first thing that came to mind. Same applies to the comment above. I can change it or you can submit a PR. Whatever you prever

simple.hs
maurges: Ok, didn't read the readme until that point. =)
`text` is sort of an stdlib because it is distributed with ghc the same way `base` is. But it sort of isn't because its version may change independently from ghc. So I'm not sure now if it's allowed.

simple.hs
maurges: Ok, I got you. =) I'll send a PR to your repo now

benhoyt:
Hi folks, thanks for the contribution. However, I believe the solution goes against the constraints ("don't read the whole file into memory") because it uses `getContents`, which reads the entire input into memory. It's also very slow (23s on my machine). And I've generally been only including solutions if they're faster than the Shell version (~15 seconds). So I appreciate the contribution, but unless those two things are fixed (in which case feel free to re-open or submit a new PR), I'll skip this version for now. Thanks anyway!

PR #68, opened 2021-03-16
Add LINQ-flavored C# version

osman-turan:
Since some entries are functional style, I would like to propose the LINQ + C# version.

Timings:

| Language  | Simple |
| --------- | ------ |
| C#        | 1.44   |
| C# (LINQ) | 7.73   |


benhoyt:
Thanks for your contribution. However, I'd like to only include one row for each language in the benchmarks table, so I'll merge this but exclude from the benchmarks. To make things "fair", I've removed the "Rust B" versions from the benchmark table too (there's actually 3 additional Rust versions anyway). But the code is all still there. CC: @BurntSushi

PR #70, opened 2021-03-16
optimized

rlonstein:
Moved & reduced the allocation, this shaves the time from 1.9s to 1s here. Now on par with optimized cpp.


benhoyt:
Thanks! Yeah, this is about twice as fast for me! I'll update the benchmarks shortly.

PR #71, opened 2021-03-17
Optimize JS further

reissbaker:
So, it turns out Node's `readline` implementation is fairly slow (at least, on my machine). The largest change in this PR can basically be summed up as:

1. Refactor the fast space-finding code out to a function that iterates over substrings terminated by a given token, `forEachTerminated`, since the previous optimized version proved that this is faster than calling `String#split` and then iterating over the returned array.
2. Read the `data` events directly from `process.stdin` rather than through the `readline` abstraction. Use `forEachTerminated` to pull out newline-terminated substrings — since `data` will be arbitrary chunks of utf-8 from stdin, and not specifically *lines* of stdin — and hand those to `handleLine`.

I also removed the class abstraction in favor of ordinary dicts with `Object.entries` for iteration, since the class didn't seem to do anything in terms of performance compared to using a dictionary of words to numbers. `Object.entries` gives any cache coherency wins and hashing-function skipping wins that you'd get from the class abstraction, and in practice it takes less code to do this with a dict of words to numbers rather than setting up the special class. The simple version has actually already switched to `Object.entries` (and got a small perf boost out of it), but it looks like the optimized version was written before that switch. There might be a minor speed boost with `Object.entries` here too due to creating less garbage for the GC to examine, not rehashing strings during `sort`, and avoiding chasing pointers to random memory locations when doing key lookups in the dict, but I didn't see a statistically significant difference on my machine when measuring that change with `hyperfine` — it just was simpler and required fewer lines of code. The only perf win I could consistently measure was from avoiding `readline`.

Overall this PR is about 15-20% faster on my machine due to avoiding `readline`, and the end result takes fewer lines of code due to removing the class abstraction!

### DIFF
--- a/optimized.js
+++ b/optimized.js
@@ -25,6 +25,7 @@ const wordHandler = word => {
 };
 
 const lineHandler = line => {
+  if(line.length === 0) return;
   wordHandler(forEachTerminated(line, " ", wordHandler));
 };
 
--- a/optimized.js
+++ b/optimized.js
@@ -1,53 +1,49 @@
 "use strict";
 
-const rd = require("readline");
-
 const dict = {};
 
-function RefNum(num) {
-  this.num = num;
-}
-
-RefNum.prototype.toString = function () {
-  return this.num.toString();
-};
-
-const wordHandler = (word) => {
-  if (word.length === 0) return;
-  word = word.toLowerCase();
-  const item = dict[word];
-  if (item !== undefined) {
-    item.num++;
-  } else {
-    dict[word] = new RefNum(1);
-  }
-};
-
-const lineHandler = (line) => {
+// Calls itemHandler on each substring terminated by the token.
+// Returns any leftover characters in the string
+const forEachTerminated = (str, token, itemHandler) => {
   let lastIndex = 0;
-  while (true) {
-    const index = line.indexOf(" ", lastIndex);
-    if (index === -1) {
-      wordHandler(line.slice(lastIndex));
-      return;
-    }
 
-    wordHandler(line.slice(lastIndex, index));
+  while (true) {
+    const index = str.indexOf(token, lastIndex);
+
+    if (index === -1) break;
+    if(index > lastIndex) itemHandler(str.slice(lastIndex, index));
+
     lastIndex = index + 1;
   }
+
+  if(lastIndex < str.length - 1) return str.slice(lastIndex);
+  return "";
+};
+
+const wordHandler = word => {
+  dict[word] = (dict[word] || 0) + 1;
+};
+
+const lineHandler = line => {
+  wordHandler(forEachTerminated(line, " ", wordHandler));
 };
 
 const endHandler = () => {
-  const keys = Object.keys(dict);
-  keys.sort((a, b) => dict[b].num - dict[a].num);
-
-  process.stdout.write(keys.map((key) => `${key} ${dict[key].num}\n`).join(""));
+  const entries = Object.entries(dict);
+  entries.sort((a, b) =>  b[1] - a[1]);
+  const output = entries.map((entry) => `${entry[0]} ${entry[1]}\n`).join("");
+  process.stdout.write(output);
 };
 
-rd.createInterface({
-  input: process.stdin,
-  output: process.stdout,
-  terminal: false,
-})
-  .on("line", lineHandler)
-  .on("close", endHandler);
+let buffer = "";
+process.stdin.setEncoding("utf-8");
+process.stdin.resume();
+
+process.stdin.on("data", (data) => {
+  buffer = forEachTerminated(buffer + data.toLowerCase(), "\n", lineHandler);
+});
+
+process.stdin.on("end", () => {
+  if(buffer.length > 0) lineHandler(buffer);
+  endHandler();
+});
--- a/optimized.js
+++ b/optimized.js
@@ -11,16 +11,16 @@ const forEachTerminated = (str, token, itemHandler) => {
     const index = str.indexOf(token, lastIndex);
 
     if (index === -1) break;
-    if(index > lastIndex) itemHandler(str.slice(lastIndex, index));
+    itemHandler(str.slice(lastIndex, index));
 
     lastIndex = index + 1;
   }
 
-  if(lastIndex < str.length - 1) return str.slice(lastIndex);
-  return "";
+  return str.slice(lastIndex);
 };
 
 const wordHandler = word => {
+  if(word.length === 0) return;
   dict[word] = (dict[word] || 0) + 1;
 };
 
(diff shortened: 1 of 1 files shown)


benhoyt:
Thanks! Sounds reasonable to me at a glance. I'll test later today and update the results.

Daninet:
Nice optimization @reissbaker. On my computer (i7-7700k, Windows 10 x64, Node.js v14.15.3) it is about 10% faster compared to the previous code.

On my computer the "RefNum" and the Object.keys() are measurably faster, I'm consistently getting about 5% improvement for that. I tried keeping the input data as a buffer and it gave me additional 5%. But I guess the performance of these microoptimizations are very dependent on the platform / CPU / OS / Node.js version.

Here is my code which gives me an additional 10% over your code:
```js
"use strict";

const dict = {};

// Calls itemHandler on each substring terminated by the token.
// Returns any leftover characters in the string
const forEachTerminated = (str, token, itemHandler) => {
  let lastIndex = 0;

  while (true) {
    const index = str.indexOf(token, lastIndex);

    if (index === -1) break;
    itemHandler(str.slice(lastIndex, index));

    lastIndex = index + 1;
  }

  return str.slice(lastIndex);
};

function RefNum(num) {
  this.num = num;
}

const wordHandler = (word) => {
  if (word.length === 0) return;
  const item = dict[word];
  if (item !== undefined) {
    item.num++;
  } else {
    dict[word] = new RefNum(1);
  }
};

const lineHandler = (line) => {
  if (line.length === 0) return;
  wordHandler(forEachTerminated(line, " ", wordHandler).toString());
};

const endHandler = () => {
  const keys = Object.keys(dict);
  keys.sort((a, b) => dict[b].num - dict[a].num);

  process.stdout.write(keys.map((key) => `${key} ${dict[key].num}\n`).join(""));
};

let buffer = Buffer.allocUnsafe(0);
process.stdin.resume();

process.stdin.on("data", (data) => {
  const str = Buffer.concat([buffer, data]).toString().toLowerCase();
  buffer = Buffer.from(forEachTerminated(str, "\n", lineHandler));
});

process.stdin.on("end", () => {
  lineHandler(buffer.toString());
  endHandler();
});
```

reissbaker:
Nice @Daninet, I'll definitely try adding the Buffer changes! I'll try re-benchmarking the class version as well. It does seem like it could be a tossup depending on Node versions and OS versions (I am in the particularly weird case of running Node on Ubuntu inside WSL2 on Windows, so there could definitely be something pretty weird about my setup); the class approach trades off writes to the dictionary with more allocations and potentially GC pressure. I would naively also expect `Object.entries` to have better cache coherency than `Object.keys` + dictionary lookups for sorting and transforming the output, but JS "arrays" are magical black boxes so that could be completely wrong.

One question: how are you measuring the changes? I use `hyperfine` with large iteration counts, and even then there's typically enough deltas between runs that it would struggle I think to conclusively decide between implementations that are 5% different.

reissbaker:
Findings for my machine, at least:

* There didn't seem to be a conclusive difference between the class version with `Object.keys` and the classless version with `Object.entries`. The classless version was slightly faster across 30 runs (by about 50ms on average)... But the error bars on that were `±  0.134 s`, so it doesn't seem statistically significant. I'd be curious to see if it's any different on @benhoyt's machine though, since he's the source of truth for the blog post. I tried a run with classes + `Object.entries` as well just to rule it out; there didn't seem to be statistically significant results with that either.
* The Buffer changes were actually about 10% worse on my machine! Intuitively that makes at least some sense to me: `Buffer.from` is presumably doing some extra copying, and I would expect setting `setEncoding('utf-8')` once would be faster than calling `.toString()` from JavaScript repeatedly — but I'm just treating V8 as a black box, and I don't know what the internals truly are. It seems like this is probably something that varies across platforms if it's significantly faster for you, since it did seem measurably slower for me.

carstenschwede:
Using `{num:1}` and `fs.readSync(process.stdin.fd)` I got ~12% faster:

Current optimized.js:
```
    Benchmark #1: node ./optimized.js < ./kjvbible_x10.txt > /dev/null
       Time (mean ± σ):      1.006 s ±  0.017 s    [User: 918.4 ms, System: 126.3 ms]
       Range (min … max):    0.982 s …  1.086 s    100 runs
```

With `new RefNum(1)`
```
    Benchmark #1: node ./optimizedRefNum.js < ./kjvbible_x10.txt > /dev/null
       Time (mean ± σ):     969.5 ms ±  36.9 ms    [User: 888.0 ms, System: 127.5 ms]
       Range (min … max):   934.3 ms … 1091.8 ms    100 runs
```

With `{num:1}`
```
    Benchmark #1: node ./optimizedObj.js < ./kjvbible_x10.txt > /dev/null
        Time (mean ± σ):     954.0 ms ±  10.7 ms    [User: 874.1 ms, System: 126.0 ms]
        Range (min … max):   938.7 ms … 989.9 ms    100 runs
```

With `fs.readSync`
```
    Benchmark #1: node ./optimizedReadSync.js < ./kjvbible_x10.txt > /dev/null
       Time (mean ± σ):     963.7 ms ±  24.4 ms    [User: 883.0 ms, System: 114.6 ms]
       Range (min … max):   941.1 ms … 1121.1 ms    100 runs
```

With `{num:1}` and `fs.readSync`
```
    Benchmark #1: node ./optimizedObjReadSync.js < ./kjvbible_x10.txt > /dev/null
        Time (mean ± σ):     886.2 ms ±  13.6 ms    [User: 807.8 ms, System: 114.2 ms]
        Range (min … max):   868.0 ms … 943.2 ms    100 runs
```

Here is the code:

```javascript
"use strict";
const fs = require("fs");
const dict = {};

// Calls itemHandler on each substring terminated by the token.
// Returns any leftover characters in the string
const forEachTerminated = (str, token, itemHandler) => {
	let lastIndex = 0;

	while (true) {
		const index = str.indexOf(token, lastIndex);

		if (index === -1) break;
		itemHandler(str.slice(lastIndex, index));

		lastIndex = index + 1;
	}

	return str.slice(lastIndex);
};

const wordHandler = (word) => {
	if (word.length === 0) return;
	const item = dict[word];
	if (item !== undefined) {
		item.num++;
	} else {
		dict[word] = {num:1};
	}
};

const lineHandler = (line) => {
	if (line.length === 0) return;
	wordHandler(forEachTerminated(line, " ", wordHandler));
};

const endHandler = () => {
	const keys = Object.keys(dict);
	keys.sort((a, b) => dict[b].num - dict[a].num);
	process.stdout.write(keys.map((key) => `${key} ${dict[key].num}\n`).join(""));
};

let buffer = "";
const BUFSIZE = 64 * 1024;
const encoding = "utf-8";
const buf = Buffer.alloc(BUFSIZE, "", encoding);
let bytesRead;
const lineToken = "\n";
for (;;) {
	bytesRead = fs.readSync(process.stdin.fd, buf, 0, BUFSIZE);
	if (bytesRead > 0) {
		buffer = forEachTerminated(buffer + buf.toString(encoding, 0, bytesRead).toLowerCase(), lineToken, lineHandler);
	} else {
		if (buffer.length > 0) lineHandler(buffer);
		endHandler();
		break;
	}
}
```

benhoyt:
@carstenschwede Thanks. If you'd like to submit a PR that makes these tweaks, feel free.

PR #72, opened 2021-03-17
 patched simple ocaml version

pkhry:


benhoyt:
This looks a bit more complex, so I'm not about updating `simple.ml`. Is this faster? In that case, what about adding it as `optimized.ml`?

pkhry:
no, its not optimized, just uses functorial interface on hashtable and doesn't use polymorphic comparison. It may be a tad bit faster. If it doesn't fit as simple.ml implementation then feel free to close the request.

mseri:
@benhoyt the specialized hash table used in this version is more verbose but also idiomatic than the previous one (and is also recommended by the stdlib documentation: the polymorphic one that was used previously is known to have many collisions).

Similarly, the use of imperative while loop (which may not do much difference here though) is a very common approach, usually recommended for fast looping due to fewer chances to allocate.

If you ask anybody in the community this would be considered a simple and immediate solution.

benhoyt:
Great, thanks -- will merge

PR #73, opened 2021-03-17
Add compile option "--gc:arc" and add optimized Nim.

tauplus:
Using `--gc:arc`  makes a bit faster.

`optimized.nim` is translated from `optimized.go`.

In my machine:
||simple|optimized|
|---|---|---|
|Go|0.99|0.38|
|Nim(original)|0.63||
|Nim|0.61|0.40|

### DIFF
--- a/benchmark.py
+++ b/benchmark.py
@@ -37,7 +37,7 @@ programs = [
     ('Swift', './simple-swift', None, 'by Daniel Muellenborn'),
     ('Perl', 'perl simple.pl', None, 'by Charles Randall'),
     ('JavaScript', 'node ./simple', 'node ./optimized', 'by Dani Biro and Flo Hinze'),
-    ('Nim', './simple-nim', None, 'by csterritt and euantorano'),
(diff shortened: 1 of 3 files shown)


benhoyt:
Thanks for the contribution -- it's definitely faster than the non-optimized version. I'll update the results in the article later today. Note that `--gc:arc` didn't work on my version of Nim (1.0.6 on Linux amd64).

tauplus:
Thank you for merging.

ARC is added in Nim 1.4.0 and stable version is 1.4.4 now.
ARC is featured function for speed up and memory efficiency.
https://nim-lang.org/blog/2020/10/15/introduction-to-arc-orc-in-nim.html

if possible, please install newest version.
https://github.com/dom96/choosenim#choosenim

benhoyt:
Thanks, updated! https://github.com/benhoyt/countwords/commit/612acdda5b9dfa77fe70097f2eb514a963767923 Will re-run the benchmarks later

PR #74, opened 2021-03-17
Add Raku simple/optmized ; test.sh and benchmark.py edited for Raku

notagoodidea:
A simple and optimized version for Raku. On my computer (i56200u, 8Go, Fedora 33, Rakudo 2021.02) is slow with the simple version around ~30s and the optimized ~16s. Maybe it is a bit too slow to be added to the benchmark script.

The optimized one do not drop to [nqp](https://github.com/Raku/nqp).

### DIFF
--- a/benchmark.py
+++ b/benchmark.py
@@ -43,6 +43,7 @@ programs = [
     ('OCaml', './simple-ml', None, 'by Nate Dobbins'),
     ('Kotlin', 'java -jar simple-kotlin.jar', None, 'by Kazik Pogoda'),
     ('Lua', 'luajit simple.lua', 'luajit optimized.lua', 'by themadsens; runs under luajit'),
+    ('Raku', 'raku simple.raku', 'raku optimized.raku', 'by notagoodidea'),
 ]
 
 times = []
--- a/optimized.raku
+++ b/optimized.raku
@@ -1,0 +1,9 @@
+my %counts is default(0);
+for open($*IN, :enc<ascii>).lines(:close) {
+  for .lc.words -> \w {
+    %counts.AT-KEY(w) == 0 ?? %counts.STORE_AT_KEY(w, 1) !! ++%counts.AT-KEY(w)
+  }
+}
+# From codesections on SO : https://stackoverflow.com/a/66500447
+sub my-reverse(\a, \b) { a.value > b.value ?? False !! True }
+%counts.sort(&my-reverse).fmt('%s',"\n").print;
--- a/simple.raku
+++ b/simple.raku
@@ -1,0 +1,3 @@
+my %counts is BagHash;
+%counts.add($_.lc.words) for $*IN.lines;
+.put for %counts.sort(*.value).reverse;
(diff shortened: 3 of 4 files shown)


benhoyt:
Yeah, I think this is too slow for inclusion at the moment. I wonder why? Perl is quite fast. Additionally, my version of Rakudo that came with `apt install rakudo` (which is fairly old, 2019.11) doesn't recognize `BagHash`. Will close for now -- feel free to reopen if there's something we're doing wrong that's easily fixable.

notagoodidea:
Thanks for the feedback !

Concerning the performance issue, the comparison with base Perl is unfair to Raku as it brings a full 00 and meta-programming system with it hence the start-up cost. A fair comparison would be to bring one of the big OO lib (Moose or other) with Perl. A member of the irc chat #raku tested quickly the script on more modern hardware (ryzen cpu 3700 or 4700) and had the time cut in half. 

The raku version in the apt repo is super old (Rakudo, the compiler/VM behind Raku is a fast moving target and performance enhancement happens at each iteration). I would advise to use [rakudo-pkg](https://nxadm.github.io/rakudo-pkg/) with up-to-date stable version of Rakudo already pre-compiled and packaged for most distro : [here a link for the instructions for Debian/Ubuntu systems](https://cloudsmith.io/~nxadm-pkgs/repos/rakudo-pkg/setup/#formats-deb).

For making it run, grabbing the latest rakudo (2021.02) from `rakudo-pkg` is the easiest fixable thing to do.

I don't think that the simple version can be improved but I can look more at the optimized version if I can do more¹. Is it fair game to drop to [nqp](https://github.com/Raku/nqp), it is a bit like using a bit of inline assembly in C (kind of)?

¹ Most of the time is passed in the AT-KEY method that extract the value from the Hash, I will keep looking how it could be possible to work around that but not sure to find something.

benhoyt:
Even at twice as fast (simple at 15 seconds) it's still almost as slow as the O(N log N) shell version, which takes 16 seconds. Can you let me know what the performance numbers are for you, for the latest version of Rakudo vs Python 3 (I'm on 3.8), for simple and optimized? That will give me a comparison point, and then I'll make a call as to whether I think it's fast enough for inclusion. If it's significantly faster than the shell version, I'll install the latest Rakudo and include it.

notagoodidea:
On Fedora 33 with Python 3.9.2 / Rakudo 2021.02, running `benchmark.py` with only `grep`, `wc`, `python` and `raku`, here are the results on my laptop : 
```
`grep`        |   0.07 |      0.06 | `grep` baseline; optimized sets `LC_ALL=C`
`wc -w`       |   0.30 |      0.41 | `wc` baseline; optimized sets `LC_ALL=C`
Python        |   2.72 |      2.01 | 
Raku          |  25.16 |     13.96 | by notagoodidea
```

benhoyt:
Thanks! I just think this is "not a good idea". :-) In other words, I think it's too slow to include in the benchmarks. It's 10x as slow as Python, and still significantly slower than the shell version. Sorry about that, and thanks anyway for the contribution.

notagoodidea:
Haha, yep, totally understand. Thanks for taking the time to review the PR and setting up this multi-language comparison!

PR #75, opened 2021-03-17
feat(zig.simple): beat all simple benchmarks

matu3ba:
NOTE: Zig std does not provide selecting a custom hasher for `StringHashMap` yet. This will likely
change after getting selfhosted.

This assumes ignoring the baseline and only compares to the other languages.

### DIFF
--- a/simple.zig
+++ b/simple.zig
@@ -1,0 +1,42 @@
+const std = @import("std");
+const WordMap = std.StringHashMap(u32);
+
+pub fn main() !void {
+    var arena = std.heap.ArenaAllocator.init(std.heap.page_allocator);
+    const ally = &arena.allocator;
+
+    var words = WordMap.init(ally);
+    const stdin = std.io.bufferedReader(std.io.getStdIn().reader()).reader();
+    var buf: [1024]u8 = undefined;
+    while (try stdin.readUntilDelimiterOrEof(&buf, '\n')) |line| {
+        for (line) |*c| c.* = std.ascii.toLower(c.*);
+        var it = std.mem.tokenize(line, " ");
+        while (it.next()) |word| {
+            const ret = try words.getOrPut(word);
+            if (ret.found_existing) {
+                ret.entry.value += 1;
+            } else {
+                ret.entry.key = try ally.dupe(u8, word);
+                ret.entry.value = 1;
+            }
+        }
+    }
+
+    const words_slice = try ally.alloc(WordMap.Entry, words.count());
+    var i: usize = 0;
+    var it = words.iterator();
+    while (it.next()) |entry| : (i += 1) {
+        words_slice[i] = entry.*;
+    }
+    std.sort.sort(WordMap.Entry, words_slice, {}, compare);
+
+    var stdout = std.io.bufferedWriter(std.io.getStdOut().writer());
+    for (words_slice) |entry| {
+        try stdout.writer().print("{s} {d}\n", .{ entry.key, entry.value });
+    }
+    try stdout.flush();
+}
+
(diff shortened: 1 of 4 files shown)


ifreund:
The Zig standard library does allow custom hash functions to be used. I chose the `StringHashMap()` for this simple solution as it is intended as the default for hashmaps with string keys. An optimized version would select the hash manually taking the specific details of this benchmark into consideration.

https://lobste.rs/s/3byl7t/performance_comparison_counting_words#c_8vbyn4
https://github.com/ziglang/zig/blob/f76bd56588e556ea580c1faa63667cc9264cc218/lib/std/hash_map.zig#L90

matu3ba:
Oh, the wording got very bad. My intention was to write that StringHashMap does not provide an interface to select the Hashing function. I added that part.

simple.zig
```diff
@@ -0,0 +1,42 @@
+const std = @import("std");
+const WordMap = std.StringHashMap(u32);
+
+pub fn main() !void {
+    var arena = std.heap.ArenaAllocator.init(std.heap.page_allocator);
```
benhoyt, line 5: Wow, this is the first time I've seen non-trivial Zig code -- allocation is explicit! I should look into this language further. Looks really useful as a C replacement for embedded development.

simple.zig
```diff
@@ -0,0 +1,42 @@
...
+    const ally = &arena.allocator;
+
+    var words = WordMap.init(ally);
+    const stdin = std.io.bufferedReader(std.io.getStdIn().reader()).reader();
+    var buf: [1024]u8 = undefined;
+    while (try stdin.readUntilDelimiterOrEof(&buf, '\n')) |line| {
+        for (line) |*c| c.* = std.ascii.toLower(c.*);
+        var it = std.mem.tokenize(line, " ");
+        while (it.next()) |word| {
+            const ret = try words.getOrPut(word);
```
benhoyt, line 15: Interesting. I presume `getOrPut` is to avoid hashing the key twice?

simple.zig
```diff
@@ -0,0 +1,42 @@
...
+                ret.entry.value = 1;
+            }
+        }
+    }
+
+    const words_slice = try ally.alloc(WordMap.Entry, words.count());
+    var i: usize = 0;
+    var it = words.iterator();
+    while (it.next()) |entry| : (i += 1) {
+        words_slice[i] = entry.*;
```
benhoyt, line 29: Out of curiosity, what does `= entry.*` mean as opposed to just `= entry`?

benhoyt:
Thanks muchly for this! I really like the looks of Zig (for the type of language it is). Merged in https://github.com/benhoyt/countwords/commit/5f93dec941ccad05d7e7426ef346f3a809a813af

benhoyt:
I'll update the benchmarks in the article later today

simple.zig
ifreund: Yep that's right. In hindsight I probably should have called the variable `entry` here instead of `ret`.

simple.zig
ifreund: `foo.*` in zig is like `*foo` in C, `.*` is a pointer dereference. In this case the `next()` function returns a `?*Entry` which is unwrapped to a `*Entry` by this `while (optional_foo) |foo|` syntax. In order to store the struct into our allocated slice by value instead of by pointer we need to deference here.

In hindsight, we could store pointers in the allocated slice for sorting instead of values (which might make things faster) as we will no longer mutate the hashmap and invalidate pointers to its entries.

pitr: In my quick test, sorting pointers ends up being 1-10% slower, I am guessing due to dereferencing cost on each compare.

PR #76, opened 2021-03-17
Added Java simple + optimized version

bit-twit:
OpenJDK11 benchmark should be ok; tested with Open JDK 11, Oracle JDK11 and graalvm-ce-java11-21.0.0.2 (uses newer JIT compiler), got mixed results.
Java incurs penalty for jvm startup and JIT warmup, so I got better results on x20 and x100 benchmarks compared to x10 benchmark and other non-native-compiled languages.

I also tried:
- AOT with GraalVM, compiling in a standalone native executable, but results were not impressive
- a more "optimized" version where I used byte[] buffer with custom code to tokenize words, pretty much same as C++ and other solutions, but it actually worked a bit slower so I left it out, seems StringTokenizer does a decent job


benhoyt:
I'm not a Java guy, but I wonder if we can simplify and remove some of the cruft (remember this is `simple.java`). What about this (see comments below):

```java
import java.io.*;
import java.util.*;

public class Code {
  public static void main(String[] args) throws java.io.IOException {
    // var reader = new BufferedReader(new InputStreamReader(System.in));
    var reader = new BufferedReader(new StringReader(" the  foo foo bar  \nthe the the the\n"));

    var counts = new HashMap<String, Integer>();
    String line;
    while ((line = reader.readLine()) != null) {
      line = line.toLowerCase();
      var words = line.split("\\s");
      for (var word : words) {
        if (!word.isEmpty()) {
          counts.put(word, counts.getOrDefault(word, 0) + 1);
        }
      }
    }

    counts.entrySet().stream()
      .sorted(Map.Entry.comparingByValue(Comparator.reverseOrder()))
      .forEach(e -> System.out.printf("%s %d\n", e.getKey(), e.getValue()));
  }
}
```

Some points to note:

* Punt on the try/catch error handling (it's "simple", after all) by just saying that main `throws java.io.IOException`
* Avoid `StringTokenzier` -- the Java docs say it's "legacy" and you should use `String.split("\\s")`, which is simpler in any case
* Use `counts.put` / `counts.getOrDefault` -- seems simpler than `counts.compute` with a lambda
* Simplify sorting using streams and `comparingByValue`
* Use `System.out.printf` instead of `PrintWriter` -- simpler and shorter

Thoughts?

bit-twit:
I cleaned up a bit and added optimized version, replacing StringTokenizer with custom buffered implementation - small difference in benchmarks.

I also added results for x20 benchmark which shows Java implementation is pretty much on par with simple C, which is expected.

I left your version as simple2.java so you can compare benchmark results - it is actually slower than other interpreted langs, which is not expected:
- regexp is actually slower than StringTokenizer, it doesn't make sense to use regular expressions just to take out spaces from input stream
- streams is slower than Collections.sort - this is actually what got me into providing a different version in the first place
- String.format actually is slower CPU+memory wise since it allocates a lot of new strings

Others:
- StringTokenizer is not officially deprecated (it should have @Deprecated annotation/tag in src/javadoc - https://docs.oracle.com/javase/7/docs/technotes/guides/javadoc/deprecation/deprecation.html)
- try{} is actually try-with-resources block, automatically closing resources allocated between ()
- I prefer lambdas, but that's more like personal taste
- "var" I would argue is not idiomatic Java code - at least not yet - and it forces you to run against JDK11, JDK8 is still widely used

benhoyt:
Thank you -- sounds good, and merged!

Yeah, I figured regexes might be slower, but when I saw the official Java docs say `StringTokenizer` was legacy and you should use this instead, it seemed weird to use it. But it doesn't seem like `StringTokenizer` is going away in any hurry.

Thanks for your contribution. I'll update the benchmarks shortly!

PR #77, opened 2021-03-17
Use bump allocator for keys

pickfire:
Should not reallocate on every keys

Before

```
  Time (mean ± σ):      60.7 ms ±  11.8 ms    [User: 53.3 ms, System: 7.5 ms]
  Range (min … max):    43.9 ms …  74.8 ms    44 runs
```

After keys arena

```
  Time (mean ± σ):      47.7 ms ±  11.5 ms    [User: 41.2 ms, System: 6.5 ms]
  Range (min … max):    37.4 ms …  70.0 ms    48 runs
 ```

After bufwriter

```
  Time (mean ± σ):      46.7 ms ±  13.5 ms    [User: 43.4 ms, System: 3.3 ms]
  Range (min … max):    29.9 ms …  63.9 ms    43 runs
```

Follow up of https://github.com/benhoyt/countwords/pull/1#issuecomment-799584798
cc @BurntSushi

### DIFF
--- a/rust/optimized/Cargo.toml
+++ b/rust/optimized/Cargo.toml
@@ -9,6 +9,7 @@ name = "countwords"
 path = "main.rs"
 
 [dependencies]
+bumpalo = "3.6.1"
 fxhash = "0.2.1"
 
 [profile.release]
--- a/rust/optimized/main.rs
+++ b/rust/optimized/main.rs
@@ -17,6 +17,7 @@ use std::io::{self, BufWriter, Read, Write};
 //
 // N.B. This crate brings in a new hashing function. We still use std's hashmap
 // implementation.
+use bumpalo::Bump;
 use fxhash::FxHashMap as HashMap;
 
 fn main() {
@@ -69,8 +70,10 @@ fn try_main() -> Result<(), Box<dyn Error>> {
     let mut ordered: Vec<_> = counts.into_iter().collect();
     ordered.sort_unstable_by_key(|&(_, count)| count);
 
+    let stdout = io::stdout();
+    let mut stdout = BufWriter::new(stdout.lock());
     for (word, count) in ordered.into_iter().rev() {
-        writeln!(io::stdout(), "{} {}", std::str::from_utf8(&word)?, count)?;
+        writeln!(stdout, "{} {}", std::str::from_utf8(&word)?, count)?;
     }
     Ok(())
 }
--- a/rust/optimized/main.rs
+++ b/rust/optimized/main.rs
@@ -6,10 +6,8 @@
 // There's nothing particularly interesting here other than swapping out std's
 // default hashing algorithm for one that isn't cryptographically secure.
 
-use std::{
-    error::Error,
-    io::{self, Read, Write},
-};
+use std::error::Error;
+use std::io::{self, BufWriter, Read, Write};
 
 // std uses a cryptographically secure hashing algorithm by default, which is
 // a bit slower. In this particular program, fxhash and fnv seem to perform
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: rust/optimized/main.rs [hunk 1]
```diff
@@ -6,6 +6,7 @@
 // There's nothing particularly interesting here other than swapping out std's
 // default hashing algorithm for one that isn't cryptographically secure.
 
+use bumpalo::Bump;
```
BurntSushi: Third party imports should be grouped together. I'd put this right above the `fxhash` comment below.

rust/optimized/Cargo.toml
```diff
@@ -9,6 +9,7 @@ name = "countwords"
 path = "main.rs"
 
 [dependencies]
+bumpalo = { version = "3.6.1", features = ["collections"] }
```
BurntSushi: Ehhhhh... I kind of think this is starting to encroach too much on the "no third party libs" rule.

I did use `fxhash`, but mostly out of convenience and because it's just a hashing function. If pressed, I could have hand-rolled an fnv hasher and it would have done just fine here I think.

But using bumpalo is pushing out a lot of complexity to a third party here. How hard is it to hand-roll a bump allocator specific to this task? Is it as easy as just creating a `Vec<u8>` at the top of `main` and passing that around?

Obviously it's @benhoyt's call here.

(All that aside, this is really cool and I'm impressed with how easy bumpalo makes this sort of optimization.)

pickfire: Err, what if I just copied bumpalo code here? Does that still considered third party? I tried doing that with `Vec` but it will encounter lifetime issue, and I can't use unsafe...

Oh, it wasn't needed by the way, the extra feature, I used it but it didn't work.

pickfire:
By the way, I used buffered writer for standard output. So it became faster again slightly. 60ms -> 50ms -> 45ms

The dependencies is just
```
├── bumpalo v3.6.1
└── fxhash v0.2.1
    └── byteorder v1.4.3
```
Still need to see whether @benhoyt allows this, bump allocator is not available in the standard library and probably cannot be written without unsafe, but it makes sense to use it here since we shouldn't allocate on every new key insert. Still, using a bump allocator is kinda cheating in a way since that will allocate a huge buffer, well GC languages does that too. So fair right?

If we want to cheat we can just do it at constant time, like https://blog.mgattozzi.dev/oxidizing-the-technical-interview/ or use ahead of time `HashMap` capacity allocation, but we didn't.

rust/optimized/Cargo.toml
pickfire: Or maybe we can just use a huge buffer, like `[u8; 1024 * 1024 * 1024]` which is an alternative to bumpalo but that requires a few additional lines of code.

rust/optimized/main.rs
pickfire: Done

BurntSushi:
Ah yeah, I see, using a `Vec<u8>` is problematic because the code wants to borrow it in the hashmap and mutably at the same time.

Dunno. I could go either way.  Like, yeah, maybe that's what GC languages do. But then again, that's their advantage. It's hard to draw the line. That's probably why it was just originally "no third party libs."

> or use ahead of time HashMap capacity allocation, but we didn't.

I tried that but couldn't see a difference. But that's a fair optimization if it matters I think.

benhoyt:
Yeah, I'd like to stick as closely to the stdlib as possible, given the constraints I set. I think it's useful to show what a language can do "out of the box". Like you say, you *could* just copy the bump allocator code in here, though it looks like a reasonable amount of code, and I don't really think this is in the spirit of the constraints anymore. So closing this one. Thanks anyway!

pickfire:
@BurntSushi I am even more surprised that I hear you are surprised by me. I thought you should have already heard of these. :D

@benhoyt Sure, let me look into that.

BurntSushi:
Yeah, I've know about bumpalo. Just never used it or happened to stumble across it solving a problem so elegantly in isolation like this.

PR #78, opened 2021-03-17
Add buffering to simple.kt

hybras:
Not all of the simple examples use buffering. However as my addition shows, buffering in kotlin is definitely simple.


benhoyt:
Note that the current code is already using `bufferedReader()`, just with the default size (usually 8KB, I think). Changing that to 64KB doesn't actually help performance noticeably (at least in my tests), so I'm going to leave as is -- it's a little bit simpler in any case.

hybras:
ok cool

PR #81, opened 2021-03-17
Fix optimized.go inconsistency

ntrrg:
Hi! I don't know if its just me, but I am getting inconsistent outputs every time I run it.

```
jiphthahel 9
jip 1
h 1
sait 1
t 1
hthahel 1
```

This version is a little bit slower, but works. And it is pretty easy to understand and maintain if I may say (more than the simple version actually :sweat_smile:).

### DIFF
--- a/optimized.go
+++ b/optimized.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"bytes"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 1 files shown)


benhoyt:
I do like this refactoring -- nice! However, I'd rather not include this change as I want the Go versions to stand pretty much as-is given they're quoted an discussed in the article.

I'm more than willing to fix bugs in it without refactoring, though I can't reproduce what you're seeing. Closing for now, but please re-open if you have a case where you can reproduce the issue you mentioned with the current `optimized.go` version.

ntrrg:
Oh yeah, that is fine, it makes sense. Let me see if I can find a way to reproduce it.

ntrrg:
So I just run this:

```shell-session
$ for _ in $(seq 10); do
  cat kjvbible_x10.txt |
    go run optimized.go |
    python3 normalize.py |
    tail

  echo
  sleep 1
done
```

And I get this:

<details>
  <summary>Output:</summary>

```
zorobabel 10
zorobabel, 10
zorobabel; 10
zuar 10
zuar, 10
zur; 10
zuriel 10
zurishaddai, 10
zuzims 10
t 1

zorobabel, 10
zorobabel; 10
zuar 10
zuar, 10
zur; 10
zuriel 10
zurishaddai, 10
zuzims 10
omen 1
w 1

zorites. 10
zorobabel 10
zorobabel, 10
zorobabel; 10
zuar 10
zuar, 10
zur; 10
zuriel 10
zurishaddai, 10
zuzims 10

zorites. 10
zorobabel 10
zorobabel, 10
zorobabel; 10
zuar 10
zuar, 10
zur; 10
zuriel 10
zurishaddai, 10
zuzims 10

zuar, 10
zur; 10
zuriel 10
zurishaddai, 10
zuzims 10
aft 1
bo 1
dy, 1
manas 1
seh, 1

zorobabel, 10
zorobabel; 10
zuar 10
zuar, 10
zur; 10
zuriel 10
zurishaddai, 10
zuzims 10
p 1
wee 1

zorobabel, 10
zorobabel; 10
zuar 10
zuar, 10
zur; 10
zuriel 10
zurishaddai, 10
zuzims 10
e 1
ther 1

zorites. 10
zorobabel 10
zorobabel, 10
zorobabel; 10
zuar 10
zuar, 10
zur; 10
zuriel 10
zurishaddai, 10
zuzims 10

zuzims 10
dr 1
fo 1
r 1
righ 1
teousness 1
tho 1
u 1
ut, 1
y, 1

zorobabel 10
zorobabel, 10
zorobabel; 10
zuar 10
zuar, 10
zur; 10
zuriel 10
zurishaddai, 10
zuzims 10
: 1
```

</details>

I am running Alpine Edge, Go 1.16. I will try on a Debian container and let you know.

benhoyt:
Wow, I see it sometimes now. So weird! I've eliminated the python3 normalize script and the tail, and can still get it to happen sometimes. So it must be a bug in the Go code. I'll investigate further. Shows the danger of optimizing without extensive testing. Thanks!

ntrrg:
I can reproduce this on a Debian Bullseye (with Linux 5.10.18, from my Alpine setup) container, but not in a Debian Stretch (with Linux 4.9.0) server, not sure if it might be any breaking change on the kernel itself.

benhoyt:
Yeah, this is really weird to me -- on my machine it's very inconsistent, and only happens 1 in every 10 or 20 times. That indicates it's a race condition (but it's simple linear, non-concurrent code) or an undefined memory issue (but Go doesn't usually have those).  I'll dig in a bit further later, but let me know if you have any ideas in the meantime.

ntrrg:
For sure, and thanks for your time! :smile:

ntrrg:
So I just removed all the unnecessary state management and it is consistent now. It is pretty much the same as the changes in this PR but without the `WordCounter` abstraction, which allows cleaning the buffer in place and has the same performance as your code. Let me know if you want another PR.

ntrrg:
As a note, the `WordCounter` abstraction (which is just ~100ms slower in my machinee) could have the same performance implementing the `io.ReaderFrom` interface.

benhoyt:
I believe the inconsistency / bug is due to the fact that I'm not handling partial reads *when there's no LF in the bytes read*. I had assumed `Read` would always give you what you asked for unless it was at the end of the file, but that's not so (the signature and semantics of `Read` are [actually quite complex](https://golang.org/pkg/io/#Reader)). For example, if for some reason a read give you `1:27 So God created ma` with the rest coming, it'll search for the last LF, not find it, and then handle the whole buffer as if it was a whole line, treating `ma` as a distinct word. In any case, I'll fix this bug in my solution in the next couple of days.

I really appreciate your contribution, and it's simpler / less bug prone than mine! However, I'd like to "own" the Go version as I coded that for and presented that in the article, and keep it fairly similar to what I had, warts and all. Thanks for helping me find this (rather tricky!) bug.

ntrrg:
Sure, I understand :smile:

The thing is, you don't need to complicate the state management that much, you can treat `'\n'` as `' '` and count words every time you find one.

https://github.com/ntrrg/countwords/blob/patch-1/optimized.go:

```go
var word []byte
buf := make([]byte, 64*1024)
counts := make(map[string]*int)
```

That is all the state I needed.

benhoyt:
Yes, that's simpler and less tricky, thank you! I think I'll basically take this approach (and add your name to the credits). Will update later today. Cheers!

ntrrg:
Don't worry, not doing it for the credits :joy: just wanted to show that even optimized Go code is easy to read and maintain :smile:

benhoyt:
Indeed!

PR #82, opened 2021-03-17
WIP: Add simple Racket version

svercl:
May not be the best code, just wanted more Lisp in here. Help is definitely wanted to improve this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@ Thanks to these contributors for additional language versions:
 * OCaml: [doesntgolf](https://github.com/doesntgolf)
 * Perl: [Charles Randall](https://github.com/charles-randall)
 * PHP: [Max Semenik](https://github.com/MaxSem)
+* Racket: [Brad Svercl](https://github.com/bradms) - not included in benchmarks as it's too slow (about 13 seconds)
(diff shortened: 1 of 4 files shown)


benhoyt:
Thanks for the contribution! However, this takes 12-13 seconds to run on my setup, so it's very slow. Is the input being buffered, or is there something else wrong here? Or is Racket just very slow. I'm disinclined to include it as is, as it just makes Racket look very slow (though maybe that's accurate!).

benhoyt:
Or I could do what I've done with Bash (takes 2+ minutes!) and merge the code but not include it in the benchmarks?

svercl:
I'm a little confused by the performance as well! It takes roughly the same time on my desktop computer. I'll admit Racket is not my lisp-like language of choice. This could definitely be improved.

svercl:
I wouldn't merge this yet until the performance is resolved.

svercl:
I did a little cleaning up, but still takes a long time. I opted not to include it in the benchmarks.

benhoyt:
Sounds good -- thanks for your contribution! (Added credits to README anyway)

PR #83, opened 2021-03-17
Improve optimize.cs: Use DictionarySlim

akraines:
Microsoft's DictionarySlim exposes a GetOrAddValueRef which saves us an indirection on increment.


benhoyt:
Thanks. This looks like a nice simplification (and very slight speedup). However, it's marked as "experimental" and it's not in the standard library, so it kind of goes against the constraints I've set. I'll skip this one for now, but thanks anyway!

PR #85, opened 2021-03-18
Add simple and optimized J implementations

jrfondren:
abbreviated benchmark results:
```
`grep`        |   0.04 |      0.04 | `grep` baseline; optimized sets `LC_ALL=C`
`wc -w`       |   0.25 |      0.36 | `wc` baseline; optimized sets `LC_ALL=C`
J             |   4.94 |      1.81 | by jrfondren 
```

Simple:
- read entire file
- replace newlines with spaces
- split by spaces to get words
- remove empty boxes created by the split of consecutive spaces
- use Key (`/.` in J, ⌸ in DyalogAPL) and Grade down to get sorted word counts

Optimized:
- allow unlimited output and use echo@sprintf to reduce I/O syscalls
- split words with a custom Sequential Machine (`;:`), a little FSM that only cares about letters vs. non-letters

### DIFF
--- a/benchmark.py
+++ b/benchmark.py
@@ -44,6 +44,7 @@ programs = [
     ('Kotlin', 'java -jar simple-kotlin.jar', None, 'by Kazik Pogoda'),
     ('Lua', 'luajit simple.lua', 'luajit optimized.lua', 'by themadsens; runs under luajit'),
     ('Zig', './simple-zig', None, 'by ifreund and matu3ba'),
+    ('J', './simple-j.ijs', './optimized-j.ijs', 'by jrfondren'),
 ]
 
 times = []
--- a/optimized-j.ijs
+++ b/optimized-j.ijs
@@ -1,0 +1,12 @@
+#! /usr/bin/env ijconsole
+require 'format/printf'
+9!:37 ] 0 _ 0 _
+m =: 256 $ 0
+m =: 1 (,(a.i.'Aa')+/i.26) }m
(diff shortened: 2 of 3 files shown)


benhoyt:
Wow, this is different, thanks! I can't read it for the life of me... :-)

Note that reading the entire file into memory is against the criteria -- you can read line-by-line or in chunks (say 64KB). Also have you tested against the required output? (see test.sh)

jrfondren:
gah, I missed that.

I didn't care too much about the exact output since it's only sorted by occurrences and there are duplicate occurrences, so sort stability would create bogus differences ... but I'm blatantly outputting the counts/words in the reverse order and I'm dropping too many characters (like punctuation).

I'll fix these.

Also, quick install instructions to try this out:
```
wget http://www.jsoftware.com/download/j903/install/j903_linux64.tar.gz
tar zxvf j903_linux64.tar.gz
j903/bin/jconsole -js "exit install 'format/printf'"
ln -s $PWD/j903/bin/jconsole ~/bin/ijconsole  # or similar
```

benhoyt:
Thanks for the installation instructions. Note that test.sh runs everything through `python3 normalize.py` when testing, to ensure words with the same frequency are sorted. And yeah, punctuation should be included in the words (per "spec").

jrfondren:
Ah, I see that now.

But, J is disqualified anyway: having to read stdin line-by-line from a script is a bridge too far without some extraordinary hacks. Thanks for the challenge :p

PR #87, opened 2021-03-18
Rust use bump allocator for keys

pickfire:
cc @BurntSushi 

Previously #77

Now this is using vec as local keys buffer instead bumpalo.

Before

```
  Time (mean ± σ):      60.7 ms ±  11.8 ms    [User: 53.3 ms, System: 7.5 ms]
  Range (min … max):    43.9 ms …  74.8 ms    44 runs
```

After keys arena

```
  Time (mean ± σ):      47.7 ms ±  11.5 ms    [User: 41.2 ms, System: 6.5 ms]
  Range (min … max):    37.4 ms …  70.0 ms    48 runs
 ```

After bufwriter

```
  Time (mean ± σ):      46.7 ms ±  13.5 ms    [User: 43.4 ms, System: 3.3 ms]
  Range (min … max):    29.9 ms …  63.9 ms    43 runs
```

With only standard library vec (without bumpalo)

```
  Time (mean ± σ):      41.8 ms ±  11.5 ms    [User: 38.6 ms, System: 3.0 ms]
  Range (min … max):    31.0 ms …  64.8 ms    69 runs
```

A single line of unsafe is required to extend the lifetime for keys which we already known. That is what the 3rd party crate does, I don't see any way around that. Even though we can do transmute without unsafe but that is slower (mean ~100ms).

### DIFF
--- a/rust/optimized/Cargo.toml
+++ b/rust/optimized/Cargo.toml
@@ -9,7 +9,6 @@ name = "countwords"
 path = "main.rs"
 
 [dependencies]
-bumpalo = "3.6.1"
 fxhash = "0.2.1"
 
 [profile.release]
--- a/rust/optimized/Cargo.toml
+++ b/rust/optimized/Cargo.toml
@@ -9,6 +9,7 @@ name = "countwords"
 path = "main.rs"
 
 [dependencies]
+bumpalo = "3.6.1"
 fxhash = "0.2.1"
 
 [profile.release]
--- a/rust/optimized/main.rs
+++ b/rust/optimized/main.rs
@@ -17,6 +17,7 @@ use std::io::{self, BufWriter, Read, Write};
 //
 // N.B. This crate brings in a new hashing function. We still use std's hashmap
 // implementation.
+use bumpalo::Bump;
 use fxhash::FxHashMap as HashMap;
 
 fn main() {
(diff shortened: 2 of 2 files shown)


BurntSushi:
@pickfire Nice! That's a good improvement. Personally, I agree with @benhoyt's "no unsafe" constraint though. Maybe this could be added as an `optimized-unsafe` Rust variant alongside the others instead. Up to @benhoyt though!

pickfire:
IMHO it isn't optimized without this, how can we keep bunch of allocations while still calling it optimized? :P Arena is one way but that requires third party crate, unsafe is another way, I don't know any other way to avoid the unnecessary allocations without unsafe or arena. I wonder how...

At the very least, we could buffer stdout for optimized. That is able to shave a few ms off although not significant. I tried looking into perf report or flamegraph, I don't understand stuff there, looks like chunks of time (~70%) just went into the hashing.

BurntSushi:
Well, there is also `optimized-customhashmap`. I think reasonable people can disagree. But basically, once you open the `unsafe` flood gates, it's tough to draw a line. On the one hand, people might look at your benchmark and go, "Well, sure, Rust is fast but look, you had to use `unsafe` to do it. Shouldn't Rust be fast without needing `unsafe` all the time?" But put it another way, you might say, "I expect Rust won't be able to be fast in every case without using `unsafe`, so I am interested in seeing how to optimize Rust programs via `unsafe`." Both are valid perspectives IMO. That's why I suggested adding it as a variant.

Anyway, benchmarks are models. All models are wrong. But some are useful. Designing them is an art-form. We could go in circles all day arguing this. At the end of the day, if we have to pick one, I think it's more useful to see how fast Rust can be without `unsafe` in the context of this benchmark.

benhoyt:
Yeah, I agree with Burntsushi here -- I think this is useful, but I want to let `optimized.rs` stand without "unsafe". Feel free to move to `optimized-unsafe`. I know it is stated a bit loosey goosey in my constraints, "*try* not to use unsafe language features".

vbarrielle:
Am I right to think this usage of unsafe is only valid if the `Vec`'s capacity is never changed? Maybe there should be such an assertion, otherwise this can be truly unsafe if the input text is in a language with a huge amount of unique words.

benhoyt:
Per discussion above, I'm going to close this for now. Feel free to reopen or submit a new PR with this as an `optimized-unsafe` version. Thanks anyway!

PR #88, opened 2021-03-18
Idiomatic optimized C++

ncm:
This also has a patch to add '-march=native' gcc and g++ compiler flags, which makes a difference. Without, gcc targets Athlon64 from 2002.

Apologies for the noise, before.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ Thanks to these contributors for additional language versions:
 * Bash: [Jesse Hathaway](https://github.com/lollipopman) - not included in benchmarks as it takes over 2 minutes
 * C#: [John Taylor](https://github.com/jftuga), [Yuriy Ostapenko](https://github.com/uncleyo) and [Osman Turan](https://github.com/osman-turan)
 * C# (LINQ): [Osman Turan](https://github.com/osman-turan) - not run in benchmarks
-* C++ optimized version: [Jussi Pakkanen](https://github.com/jpakkane) and [Adev](https://github.com/adevress)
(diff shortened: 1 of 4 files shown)


BurntSushi:
The problem with adding `-march=native` is that you probably need to make sure that's set for all other submissions too. Might be better to just not set it.

benhoyt:
Nice, thank you very much!

Off topic (from http://www.cantrip.org/cpp.html): `if new true friend not protected for explicit private union, break case and try using this`. Ha! Presumably even more so if not true friend!

benhoyt:
@BurntSushi Fair. I'm happy to let `-march=native` stand, I think -- it seems to me it's the right thing for benchmarks being run locally. I'd be happy to accept PRs for other languages that have that or equivalent flags so it's a bit more fair, though I don't have time to go through all of them right now.

BurntSushi:
@benhoyt It's just rarely used in practice, that's all. Most people get their binaries from a package manager or a github repo or whatever, and those are rarely if ever compiled using `-march=native` (or its equivalent).

benhoyt:
@BurntSushi Point taken. Yeah, I guess binaries would rather unportable if it were used heavily. I've reverted that part in https://github.com/benhoyt/countwords/commit/330e0765432b1f93a795d197a8ee308a45877dc8 -- hopefully it's more apples to apples now.

osman-turan:
@benhoyt, wasn't the custom hash table forbidden (except C)? 🤔

> Hashing: don’t roll our own hash table (with the exception of the optimized C version).

benhoyt:
@osman-turan Ah yes, that's right! I'm getting old and forgetting my own constraints. Hmmm ... I'm inclined to let it stand now that it's merged (and seeing C gets an exception)...

PR #89, opened 2021-03-19
optimized.js ~30% faster using FNV Hashing, JS Map, and toString().

peterwmwong:
Results measured on:
- MacBook Pro 2018 (2.6GHz Intel Core i7 w/16GB)
- Node v15.12.0

```
> benchmark.py # Updated to compare optimized.js original (left) and new (right)
Language      | Optimized Original | Optimized | Notes
------------- | ------------------ | --------- | -----
JavaScript    |               0.86 |      0.59 | by Dani Biro and Flo Hinze
```

I'm sure there's a few more optimizations to be had if someone inspected at the TF (JIT) generated code.

### DIFF
--- a/optimized.js
+++ b/optimized.js
@@ -1,58 +1,56 @@
 "use strict";
 const fs = require("fs");
-const dict = {};
 
-// Calls itemHandler on each substring terminated by the token.
-// Returns any leftover characters in the string
-const forEachTerminated = (str, token, itemHandler) => {
-    let lastIndex = 0;
+const BUFFER_SIZE = 64 * 1024;
+const FNV_OFFSET = 2166136261;
+const FNV_PRIME = 16777619;
 
-    for (;;) {
-        const index = str.indexOf(token, lastIndex);
-
-        if (index === -1) break;
-        itemHandler(str.slice(lastIndex, index));
-
-        lastIndex = index + 1;
+class Entry {
+    constructor(word) {
+        this.num = 1;
+        this.word = word;
     }
-
-    return str.slice(lastIndex);
-};
-
-const wordHandler = (word) => {
-    if (word.length === 0) return;
-    const item = dict[word];
-    if (item !== undefined) {
-        item.num++;
-    } else {
-        dict[word] = {num:1};
-    }
-};
-
-const lineHandler = (line) => {
-    if (line.length === 0) return;
-    wordHandler(forEachTerminated(line, " ", wordHandler));
-};
-
-const endHandler = () => {
-    const keys = Object.keys(dict);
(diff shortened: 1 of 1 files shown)


benhoyt:
Thanks for your contribution! I have allowed a custom hash in the C (and now C++) versions, but I'm loathe to allow it in the other languages. How much of the speedup is to the custom hash, and how much to the other optimizations?

peterwmwong:
@benhoyt I'd say some form of hashing is fundamental to the algorithm presented in this PR.

Hashing avoids allocating JS strings (`Buffer` -> string) **just** to increment the count of word (majority case?).  I imagine most performance-minded JS apps would employ such techniques when data originates from a `Buffer`/`TypedArrays` to avoid the need to copy/allocate.

No worries, if you'd rather not allow hashing (but only arbitrarily for C and now C++). 
It was a fun little challenge!

Thanks!

benhoyt:
Yeah, I think I'll skip it for now, but thanks -- it's good code! I may still back out the C++ custom hash (you do kinda need it in C because it doesn't have hash tables in the stdlib).

peterwmwong:
👍 

Just to be clear, maybe splitting hairs depending on your definition of a builtin hash table in a language... JS (like C) has no built-in hash/hash table (python `__hash__`, Java `hashCode`, etc). Custom hashing in JS isn't a thing as there is no builtin hashing mechanism :)

The JS spec only provides implementation guidance as :clears throat: "mechanisms that, on average, provide access times that are sublinear on the number of elements in the collection".  As per spec, JS `Map`'s lookup don't "hash" (builtin or custom) a key, but iterate through all contained entries and test for equality https://tc39.es/ecma262/#sec-map.prototype.set

Maybe one day...

benhoyt:
Oh, yeah, I guess I had assumed plain old `Object` was a hash table. It looks like it's a bit more complicated than that in most JS implementations, for example V8 uses optimized structures when small or when possible, but hash table otherwise: http://www.jayconrod.com/posts/52/a-tour-of-v8-object-representation

carstenschwede:
@peterwmwong raises an important point. `{}` has getters and setters which may look like giving you access to keys in a hash table. But these are not actual keys but object attributes of an `Object` instance, which means missing 'keys' are looked up in the prototype chain and some 'keys' are protected.

So `{}` can't be a hash table because
- a) it can't store some keys (protected attributes, e.g. `__proto__`)
- b) it is not empty on startup (e.g. `constructor`)

Both current JS implementation are thus buggy and will either fail or produce inaccurate counts if the input text contains the above examples.

This PR should be merged - JavaScript has no native hash tables.

PR #90, opened 2021-03-19
Some optimizations in c

ansingh:
I just added small optimizations in the c version to speed it up.

1. Reduced the checks required in the inner loops for detecting control / word characters.
2. Optimized allocations by keeping a linked list of buffers for words.

### DIFF
--- a/optimized.c
+++ b/optimized.c
@@ -72,64 +72,49 @@ int main() {
         return 1;
     }
 
-    char buf[BUF_SIZE];
-    int offset = 0;
+    char buf[BUF_SIZE + 3];
+    char *buf_end = buf + fread(buf, 1, BUF_SIZE , stdin);
+
     while (1) {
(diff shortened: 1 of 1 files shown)


benhoyt:
Thanks for your contribution - it definitely is about 25% faster than mine. However, I'd like to "own" the C version and keep it as is, as it's quoted and discussed in my article. In my article I acknowledge that there is much more you could do with this version, and I don't want to keep going further down that rabbit hole:

> I’m sure there’s further you could go with the C version: investigate memory-mapped I/O, avoid processing byte-at-a-time, use a fancier data structure for counting, etc. But this is quite enough for now!

Sorry about that, and thanks anyway!

ansingh:
No worries, I figured that would be the case but gave it a go anyway :)

benhoyt:
Yeah, well done on getting this faster - that's no small feat!

PR #91, opened 2021-03-19
simple.js: Shorter, faster, no collisions on reserved words

carstenschwede:
Based on the discussion in https://github.com/benhoyt/countwords/pull/89.

`{}` has protected and prepopulated attributes, which will make JS implementation fail when the input text contains such words (e.g. `__proto__` or `constructor`), which shows that `{}` is not a true hash table.

`Object.create(null)` removes the prototype chain and avoids these collisions.

### DIFF
--- a/simple.js
+++ b/simple.js
@@ -2,7 +2,8 @@ const wordCounter = Object.create(null); //{} has protected and prepopulated key
 
 require("readline").createInterface({input: process.stdin})
   .on("line", line => {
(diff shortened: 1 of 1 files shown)


benhoyt:
Thanks! I verified that the previous one did fail if the input had the word `constructor` in it (thankfully the Bible doesn't :-). And this is somewhat simpler and maybe 5% faster. I had to slightly tweak the output to avoid an extra blank line at the end, which the `normalize.py` script didn't like. Thanks again!

PR #92, opened 2021-03-20
Common Lisp simple

svercl:
This adds a version in Common Lisp. It runs in about 3-4 seconds on my machine.


benhoyt:
Thanks for the contribution! I'll add to the benchmarks shortly.

benhoyt:
Oops, accidentally closed instead of merging.

PR #93, opened 2021-03-20
Update ocaml samples

anuragsoni:
* simple.ml -> use ocaml references (mutable indirection cells) for Hashtbl value
* optimized.ml -> attempt to read upto 64kb per read
With these changes on my laptop I notice around 1.5 times improvement in the performance for `simple-ml`and 1.8 times improvement when comparing `optimized-ml` to `simple-ml`

Notes:
Testing using current stable compiler version.
4.12.0 with [flambda](https://caml.inria.fr/pub/docs/manual-ocaml/flambda.html) 

Steps to install the compiler using opam2
`opam sw create ocaml412 --packages=ocaml-variants.4.12.0+options,ocaml-options-only-flambda`


benhoyt:
Thanks for the contribution. A couple of questions:

1) My intention is for the "simple" versions to be a very simple, idiomatic way of writing it. Is what you have now more idiomatic, or is it trying to speed things up (by using the ref)? Some of the changes look simpler to me (the count words loop), other changes look more complex (moving to an array and sorting), other changes are just formatting changes. What's more idiomatic.
2) I've run the optimized version locally, and it's only about 8% faster on my machine, which I'm not sure it's worth it for the difference in complexity (I'm getting 1.2s for your updated simple version (from 1.7s), and 1.3 for the optimized).

anuragsoni:
> My intention is for the "simple" versions to be a very simple, idiomatic way of writing it. Is what you have now more idiomatic, or is it trying to speed things up (by using the ref)? Some of the changes look simpler to me (the count words loop), other changes look more complex (moving to an array and sorting), other changes are just formatting changes. What's more idiomatic.

I'd say ref's are pretty idiomatic when working on problems where one's using the mutable side of OCaml (specially when combined with loops, and other mutable datastructures like the Hash table that was already present in the simple implementation). The choice of moving to an array was made mostly because we already know the size needed for the collection ahead of time (via the hash table length). I'd say that if the size of the collection is fixed, and if the intended usage for the list was to simply collect items so we can sort then most real-world ocaml programs would end up using an array. The previous implementation was using an immutable array as a temporary store and then creating a new sorted list and consuming it right away. I don't think that will be a very common pattern in problems similar to this sample.

>     2\. I've run the optimized version locally, and it's only about 8% faster on my machine, which I'm not sure it's worth it for the difference in complexity (I'm getting 1.2s for your updated simple version (from 1.7s), and 1.3 for the optimized).

This is interesting! Is this with the compiler version I mentioned in the comment (4.12.0+flambda?) I was seeing a 0.10-0.12 second improvement over my new simple version with that. But I agree that it doesn't seem worth it to add it for the numbers you see.

anuragsoni:
I've removed optimized.ml for now.

One improvement I can think of is to list the language/compiler versions used for testing these. It could be useful to compare performance improvements with newer releases in the future :)

anuragsoni:
I've removed the switch to array so this example will be really close to the initial simple implementation, and i've also tried to keep the formatting changes to a minimum! I don't notice a big change in the version with Lists vs what I had before with Arrays, so its probably better to keep the simpler list based implementation as-is

I am not sure what OCaml version you are testing with, but if its 4.07 or higher, the last bit (converting from the hashtable to a list) can be written using the [functional iterator module ](https://caml.inria.fr/pub/docs/manual-ocaml/libref/Seq.html)added in 4.07, and that version would probably be nicer than having to write out a fold manually.

ex:
```ocaml
    List.of_seq (StringHashtbl.to_seq countwords)
    |> List.sort (fun (_, x) (_, y) -> Int.compare !y !x)
    |> List.iter (fun (w, c) -> Printf.printf "%s %d\n" w !c)
```

benhoyt:
Nice, thanks! Merging. Will update benchmarks soon.

PR #94, opened 2021-03-20
simple.py: use the generator force

jcorbin:
## This One Crazy Trick speeds up your simple Python implementation by 33% ;-)

Or at least on my machine:
- ~2.4s prior simple.py over kvj_10x
- ~1.6s this version over same
- ~1.1s optimized.py over same

So it's about half the gap to the complicated one, while begin essentially the same program, just done The One (new) way ( since like, what even is the way of Python in 2021 if not "have you tried comprehending more generators?" ;-) )

This works because it avoids N calls to counter update, instead allowing the counter constructor to count elements in one pass.

## On another note

Cheers on bringing up this lovely classic, very much appreciated!

### DIFF
--- a/simple.py
+++ b/simple.py
@@ -1,10 +1,9 @@
 import collections
 import sys
 
-counts = collections.Counter()
-for line in sys.stdin:
-    words = line.lower().split()
-    counts.update(words)
-
-for word, count in counts.most_common():
+for word, count in collections.Counter(
+    word
+    for line in sys.stdin
+    for word in line.lower().split()
+).most_common():
     print(word, count)


benhoyt:
This is very nice, and quite a nice speedup! However, I don't think it's as simple or idiomatic Python. I personally don't love nested comprehensions / generator expressions (I always forget which way around the nested loops go), and it makes the indentation odd. Additionally, I'd like to keep my Python version as it's what I discuss/quote in the article. But thanks anyway, and it's good to know how much speedup this One Crazy Track gets. :-)

jcorbin:
Heh, I mean that's why they're called "idiom"s, as they're quirks of local/personal style, not standards ( at least as the term is used classically around writing style guides, striving to eliminate the use of idiom ) ; anyhow, this was just a curiosity, and has served its true conversational meta-goal; cheers!

PR #95, opened 2021-03-20
[aside/draft] Go Parallel Version

jcorbin:
I know this is a bit out of scope with your original all-serial deal, but I'm a sucker for token-count any day of the week and couldn't resist; wanted to share where I ended up after tinkering around sporadically today on this.

The trace is now nice and dense, but presents a rather typical heap/GC scallop chart; I've not even (yet!) let myself look at an allocs profile here, but there has to be something for another day, now doesn't there 😁 

On my 12-threaded machine on the kvj_10x dataset, this goes from 0.8s to 0.2s or something like a 4x improvement; if I telescope out to a "100x" dataset, it's circa 5x better at 8.1s downto 1.5s.

I was especially delighted with how the `io.ReaderAt` / `bufio.SplitFunc` confluence came together, which was really what I wanted to share, more-so than my cute take on yet another binary merge/reduce tree ;-)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@ kjvbible_x10.txt
 cpuprofile
 simple-go
 optimized-go
+parallel-go
 simple-c
 optimized-c
 simple-cpp
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ Thanks to these contributors for additional language versions:
 * Crystal: [Andrea Manzini](https://github.com/ilmanzo)
 * D: [Ross Lonstein](https://github.com/rlonstein)
 * F#: [Yuriy Ostapenko](https://github.com/uncleyo)
-* Go: [Miguel Angel](https://github.com/ntrrg) - simplifying the Go optimized version
+* Go: [Miguel Angel](https://github.com/ntrrg) - simplifying the Go optimized version; [Joshua Corbin](https://github.com/jcorbin) - adding a [parallel Go version](https://github.com/benhoyt/countwords/blob/9db2ab6808921e649fc5212c00712e61edf6fa1c/parallel.go) for demonstration
 * Java: [Iulian Pleșoianu](https://github.com/bit-twit)
 * JavaScript: [Dani Biró](https://github.com/Daninet) and [Flo Hinze](https://github.com/laubsauger)
 * Julia: [Alessandro Melis](https://github.com/alemelis)
--- a/parallel.go
+++ b/parallel.go
@@ -1,0 +1,75 @@
+package main
+
+import (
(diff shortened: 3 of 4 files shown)


benhoyt:
This is a neat demonstration, thanks! Interestingly, on my machine it's not much faster at all, and definitely more variable. Right now I'm getting about 0.42s for both optimized-go and parallel-go (with the default proc factor) with kjvbible_x10. With a kjvbible_x100 it doesn't appear any faster for me (around the 3.7s for both). Am I holding it wrong?

In any case, I've added a link to the credits in the README (though I won't be adding this to the benchmarks) so people can find it. Thanks!

benhoyt:
BTW, I've reverted the `simple.go` change -- I'm going to keep that as per the article (it's how I'd normally write it anyway, though I'd probably put types at the top of my file).

jcorbin:
Curious how wide your machine is? (number of user-apparent processors) I'm on a recent MBP with 12 processors (Intel i7-9750H with 6 cores x2 threads).

First off, what version of Go? I'm on 1.16, and notice a significant bump even for the simple version from 1.15 to 1.16. Things like scheduler/GC interaction, let alone the never ending march of better escape analysis (less heap allocs by the compiler) can greatly affect these sorts of things. Interestingly, I seem to recall chatter about trying to make the compiler optimize patterns like `map[key]++`, but that's obviously not been done yet, since the `map[K]*V` trick was still relevant here.

My best guess is GC pressure since, in my experience at least, once you've got to through the gate of "now we're utilizing all the processors", the next big phase change is "get out of the allocs that scale with N" game, since that gets you out of the "GC rate scales with workload" game. In other words, especially for a batch workload like this, efficiency (of things like nominally background tasks) starts to become a performance (of say total run/request time) concern.

I've already got more local work that makes that phase change, was going to push to this branch is it was a draft PR ;-) ... but I'll just open a new one, and post more measurement details like I was planning to when I hung it up last night :grin:

jcorbin:
> BTW, I've reverted the `simple.go` change -- I'm going to keep that as per the article (it's how I'd normally write it anyway, though I'd probably put types at the top of my file).

Totally makes sense, I did this originally because I was just iterating in simple.go, and wanted gopls to stop complaining about the symbol collision because multiple "main" files ;-)

I very nearly reached for golang.org/x/sync/errgroup... which would've needed a go.mod... which may have tripped the worth it to turn these into subdirectories like the java/rust/etc counterparts... but I reeled back to just a `chan error` and tamped down my scope :-)

jcorbin:
Noting subsequent work in #98 fwiw

PR #96, opened 2021-03-20
Rust

pickfire:
cc @BurntSushi 
cc @vbarrielle I addressed your comments, added the assert

```
  Time (mean ± σ):      43.1 ms ±  12.9 ms    [User: 40.2 ms, System: 2.8 ms]
  Range (min … max):    30.9 ms …  63.8 ms    76 runs
```

The other commit is to use buffered stdout as well as locking stdout which will make it faster.

Follow up of #87

### DIFF
--- a/rust/optimized-unsafe/main.rs
+++ b/rust/optimized-unsafe/main.rs
@@ -86,7 +86,7 @@ fn increment<'a>(keys_outer: &'a Cell<Vec<u8>>, counts: &mut HashMap<&'a [u8], u
     assert!(keys.len() + word.len() <= keys.capacity());
     keys.extend_from_slice(word);
     // Safety: extend the lifetime of keys since are reusing the same buffer like an arena
-    let word = unsafe { std::mem::transmute(&keys[start..start + word.len()]) };
+    let word: &'a [u8] = unsafe { std::mem::transmute(&keys[start..start + word.len()]) };
     counts.insert(word, 1);
     keys_outer.set(keys); // store it back
 }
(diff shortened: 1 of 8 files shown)


rust/optimized-unsafe/main.rs
```diff
@@ -0,0 +1,92 @@
...
+    }
+    // store keys in another buffer to avoid unnecessary allocation for each key
+    let mut keys = keys_outer.take();
+    let start = keys.len();
+    // make sure if never gets larger than the keys buffer, if it reallocate the memory,
+    // the key references may be pointing to the wrong address
+    assert!(keys.len() + word.len() <= keys.capacity());
+    keys.extend_from_slice(word);
+    // Safety: extend the lifetime of keys since are reusing the same buffer like an arena
+    let word = unsafe { std::mem::transmute(&keys[start..start + word.len()]) };
```
BurntSushi: I wonder if a raw pointer cast would be better here? AIUI, there are fewer pitfalls. Or at least, annotate the receiving type to make it clearer.

pickfire: Do you mean something like this?

```suggestion
    let word: &'a [u8] = unsafe { std::mem::transmute(&keys[start..start + word.len()]) };
```

I think the `as` keyword may be even more unsafe compared to transmute when used here, since it could cast it to another type incorrectly, we are just transmuting lifetime here so I never thought about annotating types, not sure if it will be better but I am okay with both approaches.

BurntSushi: Aye. I think `&'a [u8]` would be right here?

benhoyt:
Thanks, merging!

PR #97, opened 2021-03-20
Add simple Haskell

Sacules:
By far the slowest implementation! Takes around 40s to run :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,6 @@ zig-cache
 /.idea
 java/*.class
 simple-cl
+simple-haskell
+simple.o
+simple.hi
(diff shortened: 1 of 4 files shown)


benhoyt:
Thanks for the contribution! Unfortunately I believe it goes against the constraints, as it reads the whole input into memory at once (unless I'm wrong about `getContents`), and I've generally not being merging solutions which are slower than the Shell version. If we can fix the "reading whole file into memory" issue I'd consider merging but excluding from the benchmarks (otherwise it just makes Haskell look slow :-). See also this Haskell version, which was similar https://github.com/benhoyt/countwords/pull/67#issuecomment-803044625

Sacules:
`getContents` reads the input lazily, meaning it gets one line at a time only [when needed](https://wiki.haskell.org/Haskell_IO_for_Imperative_Programmers#IO). I'll try to profile it and see if I can make it a little faster :)

benhoyt:
> getContents reads the input lazily, meaning it gets one line at a time only when needed

Oh, thank you for that clarification!

PR #98, opened 2021-03-20
parallel.go: several low hanging fruit optimizations

jcorbin:
- add standard {cpu,mem}profile flags
- eliminate dynamic allocations
  - switch to the `map[K]*V` pattern since the compiler isn't yet clever
    enough to optimize our map increments
  - process tokens purely as bytes, avoid copying across the string
    barrier
- reduce syscall rate by increasing scan buffer size from 4KB to 64KB

Comparison notes from some local runs (N=10 on the 100x dataset):
- median wall time is 65% of prior
- variability (IQR) is now 42% of prior
- prior did around 20 GCs in around 1.5s, and will continue to incur a GC every 200-300ms
- now does around 5 GCs in around 1s, and does not continue to churn its heap or GC after warmed up around the 150ms mark
- also reduced syscalls, as syscall overhead was now top in the `-cpuprofile`, with allocations out of the way
- notebook gist https://gist.github.com/jcorbin/82fb7dd9e737f6cd3b2c3a5a718f9b0d with trace screenshots in comments

### DIFF
--- a/parallel.go
+++ b/parallel.go
@@ -8,27 +8,33 @@ import (
 	"log"
 	"os"
 	"runtime"
+	"runtime/pprof"
 	"runtime/trace"
 	"sort"
-	"strings"
+	"unicode"
+	"unicode/utf8"
 )
 
 var (
 	procFactor = 1.0
 
+	bufSize      int = 64 * 1024
 	minChunkSize int = 64 * 1024
 
 	tokenSplit = bufio.ScanWords
 )
 
 func main() {
-	traceTo := flag.String("trace", "", "enable runtime tracing")
+	cpuprofile := flag.String("cpuprofile", "", "write cpu profile to `file`")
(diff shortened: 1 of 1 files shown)


jcorbin:
Cross-linking trace screenshot comments for convenience, as they're a bit more telling than the notebook stats in the gist itself:
- https://gist.github.com/jcorbin/82fb7dd9e737f6cd3b2c3a5a718f9b0d#gistcomment-3673292
- https://gist.github.com/jcorbin/82fb7dd9e737f6cd3b2c3a5a718f9b0d#gistcomment-3673296

benhoyt:
You're having fun! Sure, I'll merge.

benhoyt:
Nice work, by the way. Yeah, I thought it was strange there wasn't an in-place version of `bytes.ToLower` already -- usually the stdlib is pretty good about offering those kinds of things.

PR #100, opened 2021-03-21
Add optimized zig version

ansingh:
I've attempted to optimize the simple.zig by reading directly into a buffer and parsing the words manually. Also selected a different hash function for the map.

No unsafe usage, no custom hash implementation and it seems to run pretty fast (faster than optimized c, at least on my machine).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ optimized.cmx
 optimized.o
 simple-kotlin.jar
 simple-zig
+optimized-zig
 zig-cache
 /.idea
 java/*.class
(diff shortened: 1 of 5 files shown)


benhoyt:
Nice, thank you! On my machine I get the optimized C as 0.24s and the optimized Zig at 0.25s across many runs (but it's very close!). I've merged and updated the benchmarks.

ansingh:
I guess the difference is due to some slower library call implementation on macos. I seem to get results similar to yours on linux but under osx, zig is significantly faster.

Mac : optimized C is 0.22 and zig is 0.19
Linux : optimized C is 0.18-0.19 and zig is 0.19

PR #102, opened 2021-03-22
c++ version faster, more robust

ncm:
Half the speedup comes from to_lower.


benhoyt:
I'm not opposed to merging this, but I'm not seeing any significant speed difference on my machine.

ncm:
"Faster on some hardware"? Anyway, with asserts against overrun.
Curiously, when I build with "-march=native" or -O3 it runs 10ms slower.

PR #103, opened 2021-03-23
Add simple Pascal version

osman-turan:
I have implemented the simple Pascal version for Free Pascal with Delphi dialect. I did not check if it's compatible with the original Delphi.

| Language | Simple | Optimized |
| -------- | ------ | --------- |
| C        | 1.09   | 0.24      |
| Pascal   | 3.86   |           |

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,4 @@ zig-cache
 /.idea
 java/*.class
 simple-cl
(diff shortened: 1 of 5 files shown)


benhoyt:
Nice work, thanks! I'll update the benchmarks a bit later today.

PR #105, opened 2021-03-23
Optimized zig - Pre ensure buffer capacity for the hash map, and chosen a simpler hash

ansingh:
Small optimizations in the optimized zig version I submitted.
Allocated the hash map in advance and switched to the same hash the c version uses.


benhoyt:
Thanks! I'll update the benchmarks later today.

PR #107, opened 2021-03-23
Add simple Haxe version

osman-turan:
I have implemented the simple Haxe version which runs with `-interp` flag. So it's not compiled to C++ or any other target. This is not ideal performance-wise. But I think it fits in the "simple" usage.

| Language   | Simple | Optimized |
| ---------- | ------ | --------- |
| JavaScript | 2.01   | 1.15      |
| Haxe       | 19.00  |           |

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@ Thanks to these contributors for additional language versions:
 * D: [Ross Lonstein](https://github.com/rlonstein)
 * F#: [Yuriy Ostapenko](https://github.com/uncleyo)
 * Go: [Miguel Angel](https://github.com/ntrrg) - simplifying the Go optimized version; [Joshua Corbin](https://github.com/jcorbin) - adding a [parallel Go version](https://github.com/benhoyt/countwords/blob/9db2ab6808921e649fc5212c00712e61edf6fa1c/parallel.go) for demonstration
+* Haxe: [Osman Turan](https://github.com/osman-turan)
 * Java: [Iulian Pleșoianu](https://github.com/bit-twit)
 * JavaScript: [Dani Biró](https://github.com/Daninet) and [Flo Hinze](https://github.com/laubsauger)
 * Julia: [Alessandro Melis](https://github.com/alemelis)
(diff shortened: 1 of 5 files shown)


benhoyt:
Appreciate your contributions. I'm not sure about merging this one: 1) I don't know too much about Haxe, but it seems to me it's fairly specialized compared to PHP, and that there's not much point in using it over PHP if you're going to use `-interp`? 2) It's very slow at the moment, slower than the Shell. I haven't included any other languages in the benchmark if they're slower than the shell version.

BTW, just so you know, I'll probably set myself a date (maybe next week) after which I'll move the countwords repository to archived. I've been struggling to keep up with all the amazing contributions! I think what we have there now is a pretty good list.

osman-turan:
> 1) I don't know too much about Haxe, but it seems to me it's fairly specialized compared to PHP, and that there's not much point in using it over PHP if you're going to use -interp?

Haxe is an open-source successor of Adobe's ActionScript 2. It can target various languages (C++, PHP, JavaScript, etc.) for source codes and VMs (.NET Framework, JVM, Neko, etc.) for byte codes. Since I understand the goal of "simple versions" is to reflect daily usage without heavy optimization I decided to use Haxe's interpreter without any compilation step. I can configure to use C++ target if you want (you will need 2 extra steps for the first usage in your test environment). But Haxe's strong point is not performance but rather cross-platform development with a single code base. So I thought it should fit in your goals (excluding performance point).

> 2) It's very slow at the moment, slower than the Shell. I haven't included any other languages in the benchmark if they're slower than the shell version.

I have tried C++ target at my machine (it outputs a C++ project and then compiles it to native executable with available C++ compiler):

| Language   | Simple | Optimized |
| ---------- | ------ | --------- |
| JavaScript | 2.01   | 1.15      |
| Haxe       | 19.00  |           |
| Haxe (C++) | 11.21  |           |

Please feel free to reject the PR if you think it does not meet to your criterias.

benhoyt:
Ah, sorry, I was getting confused with Facebook's "Hack" language (basically PHP but with static typing). I think I will decline this one due to the slow speed and Haxe not being super widely used/known. Sorry!

PR #108, opened 2021-03-24
fix: use latest OCaml compiler

rbjorklin:
Not sure how open you are to using docker but this will use a compiler [version that's almost 3 years newer](https://ocaml.org/releases/index.html) and produces a ~20% improvement on my machine.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -184,7 +184,7 @@ php simple.php <kjvbible_x10.txt | python3 normalize.py >output.txt
 git diff --exit-code output.txt
 
 echo OCaml simple
-ocamlopt.opt -O3 simple.ml -o simple-ml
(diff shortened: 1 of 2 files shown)


benhoyt:
Thanks, but I think I'm going to skip this. I'd rather not use Docker for this, and (for better or worse) my gcc version and other compilers I've used "apt install" for are probably a bit old too. I thought about building from source, but I only have a limited amount of time to spend on this, so I'm going to pass. Thanks for your contribution!

rbjorklin:
Thanks for considering it. Would you be open to something like this for the compiler installation?

`apt-get install -y opam ; opam init --compiler ocaml-variants.4.11.2+flambda ; eval $(opam env)`

EDIT: Just saw your note about no longer accepting contributions. You can disregard this, thanks for your time!

PR #109, opened 2021-03-26
faster c++

ncm:
Dunno if this counts as a "new submission".

But on my hardware (skylake) it now shows up as exactly as fast as the C code.
Curious about yours.

### DIFF
--- a/optimized.cpp
+++ b/optimized.cpp
@@ -54,10 +54,11 @@ table<(1<<16), (1<<19)> counts;
 int main() {
   std::ios::sync_with_stdio(false);
   auto to_lower = [](unsigned char c) { return c | (-(c-'A' < 26) & '\x20'); };
-  for (std::istreambuf_iterator<char> it(std::cin), end; it != end; ++it) {
-    if (*it <= ' ') {
+  std::streambuf& in = *std::cin.rdbuf();
+  for (int c; (c = in.sbumpc()) != EOF;) {
+    if (unsigned(c) <= ' ') {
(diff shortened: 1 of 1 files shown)


benhoyt:
It doesn't count as a new submission :-), but on my machine it's no better, perhaps slightly worse. Before:

```
C             |   1.07 |      0.24 | 
C++           |   1.90 |      0.28 | optimized by Jussi P, Adev, Nathan M
```

After:

```
C             |   1.01 |      0.24 | 
C++           |   1.88 |      0.29 | optimized by Jussi P, Adev, Nathan M
```

Going to skip this one. Thanks again!

ncm:
Getting more curious about what you are running these benchmarks on, and why the results are so different from mine.

benhoyt:
Sure: it's a Dell XPS 15 laptop with an i7-6700HQ CPU (2.6GHz, 8 cores), 32GB RAM. I'm running Ubuntu Linux 20.04, kernel version 5.8.0.

ncm:
That is stranger still, because I, also, am benchmarking on a 6700HQ, reporting 2.6 GHz.

Compiling with Gcc release 10.2.

benhoyt:
Interesting coincidence! One difference here is our gcc version: I'm on 9.3.0, which came with `apt` on my Ubuntu version.

ncm:
It seems to me like you should be worried to see so much variance in performance vs. an identical target CPU. Gcc 9 and gcc 10 code generators are not very different.

PR #111, opened 2021-03-31
Fix optimized.c wrong output after remove end-line

nploi:
- When i remove end-line of `small.txt` 
- then run `gcc -O2 optimized.c -o optimized-c && ./optimized-c < small.txt`
- output is wrong

Before:
```
bar 4
the 2
foozle 1
```

After fixed:
```
bar 5
the 2
foozle 1
```


benhoyt:
Hi @nploi -- thanks for pointing this out. The bug is definitely there, however, I don't think your fix is correct. `'\n'` is already `<= ' '` (`'\n'` is 10 and `' '` is 32, so it's less than) so that change doesn't actually do anything. However, I've fixed in https://github.com/benhoyt/countwords/commit/16ef3b62f116c1373238cb49f4a4935cee5ea3c0

PR #112, opened 2021-04-20
init

why-not-try-calmer:
Second commits handles upper to lowercase strings.

Signed-off-by: Nycticorax <mrnycticorax@gmail.com>


benhoyt:
Thanks for this version! Looks nice and simple. However, I'm not taking new submissions anymore - sorry about that! See https://github.com/benhoyt/countwords#note

why-not-try-calmer:
I understand that you don't want to spend more time on this, but in my humble opinion, Haskell is a venerable language and not including it is somewhat underwhelming for a project aiming at cross-languages comparisons. Hence this proposal: How about I add all the footwork for running the test(s) in another PR? That would leave you with 0 work to do, right?

benhoyt:
Fair enough. :-) I'll add it if 1) the benchmark runs somewhat faster than the Shell version (say under 10 seconds), and 2) if you add the updates to benchmark.sh, test.sh, and the credits in README.md. Thanks!

why-not-try-calmer:
> Fair enough. :-) I'll add it if 1) the benchmark runs somewhat faster than the Shell version (say under 10 seconds), and 2) if you add the updates to benchmark.sh, test.sh, and the credits in README.md. Thanks!

Sounds like a deal. I'll need a couple of days before getting back to this but things will be made as you specified.

Thanks for keeping an open mind on this, and see you soon.

why-not-try-calmer:
I think I am barely under the upper bound you required:
Screenshot_20210425_203503
The handling of the build and executable is a little bit noisy, with a `cd` and a `cd` out. Couldn't figure out how to build into the `haskell` directory from parent directory, but I may find out. Let me know what you prefer!

benhoyt:
Hi @why-not-try-calmer - unfortunately on my machine it takes 18-20 seconds to run, and runs all cores at 70-100% CPU (most of the other solutions are single threaded and only use one CPU). So this is doing an awful lot more work (on the order of 100x) than the other solutions.

If you want, I could include the code and link it [here](https://benhoyt.com/writings/count-words/#other-languages) but not include the results in the benchmark. I've done that for a couple of other solutions, including Bash.

However, in my mind -- unless we're doing something fundamentally wrong, it just makes Haskell look like it's really not designed for this sort of problem? Are you using the best data structures? Is there something wrong with the parallelization that's actually making it take longer than a simple single-threaded solution? (I don't know Haskell at all.)

why-not-try-calmer:
> Hi @why-not-try-calmer - unfortunately on my machine it takes 18-20 seconds to run, and runs all cores at 70-100% CPU (most of the other solutions are single threaded and only use one CPU). So this is doing an awful lot more work (on the order of 100x) than the other solutions.
> 
> If you want, I could include the code and link it [here](https://benhoyt.com/writings/count-words/#other-languages) but not include the results in the benchmark. I've done that for a couple of other solutions, including Bash.
> 
> However, in my mind -- unless we're doing something fundamentally wrong, it just makes Haskell look like it's really not designed for this sort of problem? Are you using the best data structures? Is there something wrong with the parallelization that's actually making it take longer than a simple single-threaded solution? (I don't know Haskell at all.)

Hey Ben, thanks for your reply. I am aware it's not optimized at all, and I thought that this was OK since on my machine it's running slightly under the threshold of 10s and also because I felt you wanted to create contrast between the `simple` and the `optimized` versions. My understanding was indeed that the former should prioritize legitibility and simplicity over the latter, so it's in this spirit that I am submitting `simple.hs`, which as you rightly notice, makes no intelligent use of parallelization at all (the runtime is trying to spread exectution over CPUs but of course it cannot do miracles without proper design).

I should be able to  submit an optimized version soon enough though, one that should run closer to your lower-bound requirements. You'll tell me what to do with the simple version then :)

benhoyt:
Yeah, what you're saying is true -- I do want to create a contrast between simple and optimized. But when the simple is so slow to start with compared to the other languages it doesn't give a good impression of the language (and it also takes a long time to run the benchmarks).

why-not-try-calmer:
Agreed. Also this little game is an interesting challenge for Haskell, which has lazy evaluation and immutability by default. You can imagine the kind of friction it may cause for in-place mutability... I should have identified this earlier and used a hash table. Will do ASAP.

why-not-try-calmer:
After numous attempts trying to beat the timer, I discussed with a friend who happens to do much more work with IO and texts than I do, and he pointed me to the extremely simple but sound implementation at https://rosettacode.org/wiki/Word_frequency#Haskell. I drew heavily on it for cooking my thing.

benhoyt:
This definitely looks simpler, nice work. However, I'm still not sure I want to include it, it takes 17-18 seconds on my machine, which is significantly longer than the very slow Shell implementation (already at the bottom of the list). So it'll mean Haskell just sits there right at the bottom of the list. Maybe that's okay, and just confirms that Haskell is not the right tool for this job?

Also, I believe there's an extra blank line at the end of the output, which causes `normalize.py` to choke (I manually deleted that line and the output matched).

why-not-try-calmer:
Hey again Ben, last commit removed the extra line break mentioned in your last message.

As for publishing my version, here's how I see it. Either:

a) the version proposed here is the absolute best that Haskell can do. Then Haskell is painfully slow for this type of work and there is no reason to hide it, perhaps provided a one-sentence disclaimer warning the visitor that (1) Haskell is not fast at printing and correlatedly that (2) a functionnally identical program without the printing requirement (perhaps substituting writing to disk or some other I/O task) would have Haskell fare much better; 

or much more plausibly:

b) the version proposed here can be improved by a significant margin, but since that consitutes an interesting challenge and since any challenge needs a baseline, my version can be that baseline.

I know for a fact that Haskell sucks at printing, even though it does not suck at all for I/O things in general -- it just requires some footwork for I/O things. If I am correct then the truth is some way between (a) and (b) and exposing this version is setting a challenge for the next guy around to help us pinpoint the truth between (a) and (b)...

why-not-try-calmer:
As for where exactly you want to mention this version (main list or "other languages") I don't really mind either.

benhoyt:
Fair enough, thanks for pursuing this! I've just merged this PR (and added a few tweaks in https://github.com/benhoyt/countwords/commit/464d74ee6c45f54a0394d5862ad1d213035bb618). I've updated the credits/results in the article too. It ends up being just above the Shell version, so at least not right at the bottom. :-)

why-not-try-calmer:
> Fair enough, thanks for pursuing this! I've just merged this PR (and added a few tweaks in [464d74e](https://github.com/benhoyt/countwords/commit/464d74ee6c45f54a0394d5862ad1d213035bb618)). I've updated the credits/results in the article too. It ends up being just above the Shell version, so at least not right at the bottom. :-)

Thanks for your trust Ben, it's been a pleasure. And rest assured that I'll rush to you again if I come across The Absolute Best Way ©™®

PR #113, opened 2021-06-24
Further optimize Go implementation (~ 10% improvement)

kevburnsjr:
This updated `optimized.go` is about 10% more efficient than the existing implementation.

- Reduce system calls using buffered output writer
- Reduce allocations by replacing `int` pointer with `struct` pointer
- Reduce data copying by removing `word` buffer, performing lowercase transformation against read buffer directly

### DIFF
--- a/optimized.go
+++ b/optimized.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bufio"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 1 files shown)


kevburnsjr:
## Before
```
> time ./optimized-go <kjvbible_x10.txt > out.txt

real    0m1.385s
user    0m1.304s
sys     0m0.087s
```
## After
```
> time ./optimized-go <kjvbible_x10.txt > out.txt

real    0m1.225s
user    0m1.215s
sys     0m0.019s
```

benhoyt:
Thanks! On my testing I get about a 4.5% improvement (I'm running under benchmark.py so it does 5 runs and takes the min time -- this also outputs to `>/dev/null`, same as all the other benchmarks). I was curious which of the optimizations is actually helping most, and I get *all* of the improvement from just the `bufio` outputting change. In my tests the other two changes don't actually help a noticeable amount. Is this the case for you too?

kevburnsjr:
@benhoyt It makes sense that the total reduction would appear less when running normalize since its virtually constant execution time dilutes the performance improvements. I did notice an improvement from the other changes. 🤷‍♂️

benhoyt:
Thanks again @kevburnsjr. I'm going to leave the repo and blog entry stand as is, but I do think the bufio outputting change is a good one.

PR #114, opened 2022-03-01
Using a larger write buffer (BufWriter) is 25% faster

kimono-koans:


BurntSushi:
Could you please show the precise commands you ran to execute the benchmark that said this resulted in a 25% boost?

benhoyt:
Is this increasing the buffer size, or changing from no buffer to a buffer? It looks like the latter, but I don't know the Rust stdlib.

kimono-koans:
> Is this increasing the buffer size, or changing from no buffer to a buffer? It looks like the latter, but I don't know the Rust stdlib.

writeln! macro docs seems to says writeln! writes to a buffer, but doesn't say anything about it -- "Writes formatted data into a buffer".  See: https://doc.rust-lang.org/std/macro.write.html  I know the BufWriter standard buffer is 8K.  I had initially presumed as well that it was a change from no buffer to a buffer.

kimono-koans:
> Could you please show the precise commands you ran to execute the benchmark that said this resulted in a 25% boost?

Big fan of yours!  Let's see:

```
hyperfine -w 3 "cat ../kjv* | ./optimized-personal/target/release/countwords" "cat ../kjv* | ./optimized/target/release/countwords"
Benchmark 1: cat ../kjv* | ./optimized-personal/target/release/countwords
  Time (mean ± σ):      27.5 ms ±   0.3 ms    [User: 26.4 ms, System: 3.6 ms]
  Range (min … max):    27.0 ms …  28.3 ms    101 runs

Benchmark 2: cat ../kjv* | ./optimized/target/release/countwords
  Time (mean ± σ):      36.3 ms ±   0.3 ms    [User: 30.6 ms, System: 6.8 ms]
  Range (min … max):    35.7 ms …  37.2 ms    77 runs

Summary
  'cat ../kjv* | ./optimized-personal/target/release/countwords' ran
    1.32 ± 0.02 times faster than 'cat ../kjv* | ./optimized/target/release/countwords'  
```

BurntSushi:
@kimono-koans That's not the right benchmark. You need to test it on `kjvbible_x10.txt` (which isn't part of the repo, but is instead generated from `kjvbibel.txt`). See `benchmark.py`. `kjvbible_x10.txt` is generated by `test.sh` with the following command:

```
$ cat kjvbible.txt kjvbible.txt kjvbible.txt kjvbible.txt kjvbible.txt kjvbible.txt kjvbible.txt kjvbible.txt kjvbible.txt kjvbible.txt >kjvbible_x10.txt
```

While this change might reflect a 10ms improvement, it's unlikely that this change reflects such a sizable performance improvement regardless of input, because that 10ms isn't going to improve proportionally.

kimono-koans:
> While this change might reflect a 10ms improvement, it's unlikely that this change reflects such a sizable performance improvement regardless of input, because that 10ms isn't going to improve proportionally.

You're right.  Much more modest gain.

```
 hyperfine -w 3 "cat ../kjvbible_x10.txt | ./optimized-personal/target/release/countwords" "cat ../kjvbible_x10.txt | ./optimized/target/release/countwords"
Benchmark 1: cat ../kjvbible_x10.txt | ./optimized-personal/target/release/countwords
  Time (mean ± σ):     218.7 ms ±   0.9 ms    [User: 214.9 ms, System: 10.7 ms]
  Range (min … max):   216.9 ms … 220.4 ms    13 runs

Benchmark 2: cat ../kjvbible_x10.txt | ./optimized/target/release/countwords
  Time (mean ± σ):     225.2 ms ±   0.8 ms    [User: 216.5 ms, System: 15.6 ms]
  Range (min … max):   223.9 ms … 226.5 ms    13 runs

Summary
  'cat ../kjvbible_x10.txt | ./optimized-personal/target/release/countwords' ran
    1.03 ± 0.01 times faster than 'cat ../kjvbible_x10.txt | ./optimized/target/release/countwords'
```

BurntSushi:
Yeah that's why I didn't use a buffer in the first place. Wasn't worth the increase in code complexity IMO.

BurntSushi:
Sorry, to be clear, `writeln!` is using a buffer. But it is a line-based buffer. However, since we're writing a line at a time, I would guess that the writes are actually ultimately unbuffered.

The main reason why buffering doesn't help so much when you crank up the input is that the number of lines written is equal to the number of _unique_ words in the input. As you add more input, the rate of growth in the number of unique words slows. And thus, so to does the perf benefit of using buffering. (Or, really, optimizing anything about the actual output part of the program.)

kimono-koans:
That makes sense!

benhoyt:
Given the above discussion, I think I'll close this. Thanks anyway @kimono-koans

